PR #2, opened 2016-06-07
Bson streambuf

rkargon:
The commit history seems to have been messed up when I pulled changes to master that included working CMake files. 

The only relevant files are `src/bson_mapper/bson_streambuf.*pp`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@
 
 # OSX
 .DS_Store
+
+# Editor-specific settings
+.clang_complete
--- a/.gitignore
+++ b/.gitignore
@@ -12,12 +12,3 @@
 
 # Editor-specific settings
 .clang_complete
-
-# CMake files
-CMakeCache.txt
-CMakeFiles
-CMakeScripts
-Makefile
-cmake_install.cmake
-install_manifest.txt
-CTestTestfile.cmake
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,12 @@
 
 # Editor-specific settings
 .clang_complete
+
+# CMake files
+CMakeCache.txt
+CMakeFiles
+CMakeScripts
+Makefile
+cmake_install.cmake
+install_manifest.txt
+CTestTestfile.cmake
--- a/src/bson_mapper/bson_streambuf.cpp
+++ b/src/bson_mapper/bson_streambuf.cpp
@@ -27,7 +27,7 @@
 namespace bson_mapper {
 BSON_MAPPER_INLINE_NAMESPACE_BEGIN
 
-bson_output_streambuf::bson_output_streambuf(document_callback *cb)
+bson_output_streambuf::bson_output_streambuf(document_callback cb)
     : _cb(cb), _data(nullptr, [](uint8_t *p) {}), _bytes_read(0), _len(0) {
 }
 
@@ -68,7 +68,7 @@ int bson_output_streambuf::insert(int ch) {
 
     // This creates the document from the given bytes, and calls the user-provided callback.
     if (_bytes_read == _len) {
-        (*_cb)({std::move(_data), _len});
+        _cb({std::move(_data), _len});
         _bytes_read = 0;
         _len = 0;
     } else if (_bytes_read > _len && _len >= 4) {
--- a/src/bson_mapper/bson_streambuf.cpp
+++ b/src/bson_mapper/bson_streambuf.cpp
@@ -106,7 +106,7 @@ int char_array_streambuf::pbackfail(int ch) {
 }
 
 std::streamsize char_array_streambuf::showmanyc() {
-    return _end - _current;
+    return std::distance(_current, _end);
 }
 
 std::streampos char_array_streambuf::seekpos(std::streampos sp, std::ios_base::openmode which) {
@@ -139,7 +139,7 @@ std::streampos char_array_streambuf::seekoff(std::streamoff off, std::ios_base::
 }
 
 bson_input_streambuf::bson_input_streambuf(const bsoncxx::document::view &v)
-    : char_array_streambuf((const char *)v.data(), v.length()) {
+    : char_array_streambuf(reinterpret_cast<const char *>(v.data()), v.length()) {
 }
 
 BSON_MAPPER_INLINE_NAMESPACE_END
--- a/src/bson_mapper/bson_streambuf.cpp
+++ b/src/bson_mapper/bson_streambuf.cpp
@@ -77,5 +77,70 @@ int bson_output_streambuf::insert(int ch) {
     return ch;
 }
 
+char_array_streambuf::char_array_streambuf(const char *data, size_t len)
+    : _begin(data), _end(data + len), _current(_begin) {
+}
+
+int char_array_streambuf::underflow() {
+    if (_current == _end) {
+        return EOF;
+    }
+    return *_current;
+}
+
+int char_array_streambuf::uflow() {
+    if (_current == _end) {
+        return EOF;
+    }
+    return *(_current++);
+}
+
+int char_array_streambuf::pbackfail(int ch) {
+    // We can't put back a character if the current pointer is all the way at the beginning,
+    // or if the given character does not match
+    if (_current == _begin || (ch != EOF && ch != _current[-1])) {
+        return EOF;
+    }
+    _current--;
+    return *_current;
+}
+
+std::streamsize char_array_streambuf::showmanyc() {
+    return _end - _current;
+}
+
+std::streampos char_array_streambuf::seekpos(std::streampos sp, std::ios_base::openmode which) {
+    if (which & std::ios_base::in) {
+        _current = std::max(std::min(_begin + sp, _end), _begin);
+    }
+    // clamp current pointer to be within the buffer.
+    return std::distance(_current, _end);
+}
+
+std::streampos char_array_streambuf::seekoff(std::streamoff off, std::ios_base::seekdir way,
+                                             std::ios_base::openmode which) {
+    // offset is relative to either the beginning, current, or end pointers.
+    if (which & std::ios_base::in) {
+        switch (way) {
+            case std::ios_base::beg:
+                _current = _begin + off;
+                break;
+            case std::ios_base::cur:
+                _current = _current + off;
+                break;
+            case std::ios_base::end:
+                _current = _end + off;
+                break;
+        }
+    }
+    // clamp current pointer to be within the buffer.
+    _current = std::max(std::min(_current, _end), _begin);
+    return std::distance(_begin, _current);
+}
+
+bson_input_streambuf::bson_input_streambuf(const bsoncxx::document::view &v)
+    : char_array_streambuf((const char *)v.data(), v.length()) {
+}
+
 BSON_MAPPER_INLINE_NAMESPACE_END
 }  // namespace bson_mapper
--- a/src/bson_mapper/bson_streambuf.cpp
+++ b/src/bson_mapper/bson_streambuf.cpp
@@ -1,0 +1,62 @@
+#include <cassert>
+#include <cstdio>
+#include <iostream>
+#include <streambuf>
+
+#include <mongocxx/client.hpp>
+#include <mongocxx/instance.hpp>
+
+#include "bson_streambuf.hpp"
+
+bson_output_streambuf::bson_output_streambuf(mongocxx::collection *coll){
+    this->coll = coll;
+}
+
+bson_output_streambuf::~bson_output_streambuf(){
+    if (data != nullptr){
+        delete[] data;
+    }
+}
+
+int bson_output_streambuf::underflow(){
+    return EOF;
+}
+
+int bson_output_streambuf::overflow(int ch){
+    int result = EOF;
+    assert(ch >= 0 && ch <= UCHAR_MAX);
+    result = insert(ch);
+    return result;
+}
+
+int bson_output_streambuf::insert(int ch){
+    bytes_read++;
+
+    // for first four bytes, build int32 that contains document size
+    if (bytes_read <= 4){
+        len += (ch << (8 * (bytes_read-1)));
+    }
+    // once doc size is received, allocate space.
+    if (bytes_read == 4){
+        // TODO perhaps fail unless 0 < len <= MAX_BSON_SIZE?
+        data = new uint8_t[len];
+        *(size_t *)data = len;
+    }
+
+    if (bytes_read > 4){
+        data[bytes_read-1] = (uint8_t)ch;
+    }
+
+    if (bytes_read == len){
+        // insert document, reset data and length
+        bsoncxx::document::view v(data, len);
+        coll->insert_one(v);
+        delete[] data;
+        data = nullptr;
+        bytes_read = 0;
+        len = 0;
+    } else if (bytes_read > len && len >= 4){
+        return EOF;
+    }
+    return ch;
+}
--- a/src/bson_mapper/bson_streambuf.cpp
+++ b/src/bson_mapper/bson_streambuf.cpp
@@ -29,12 +29,6 @@ BSON_MAPPER_INLINE_NAMESPACE_BEGIN
 bson_output_streambuf::bson_output_streambuf(mongocxx::collection coll) : coll(coll) {
 }
 
-bson_output_streambuf::~bson_output_streambuf() {
-    if (data != nullptr) {
-        delete[] data;
-    }
-}
-
 int bson_output_streambuf::underflow() {
     return EOF;
 }
@@ -61,8 +55,8 @@ int bson_output_streambuf::insert(int ch) {
         if (len > BSON_MAX_SIZE) {
             throw std::invalid_argument("BSON document length is too large.");
         }
-        data = new uint8_t[len];
-        std::memcpy(data, &len, 4);
+        data = std::unique_ptr<uint8_t[]>(new uint8_t[len]);
+        std::memcpy(data.get(), &len, 4);
     }
 
     if (bytes_read > 4) {
@@ -71,10 +65,8 @@ int bson_output_streambuf::insert(int ch) {
 
     if (bytes_read == len) {
         // This inserts the document into the collection, and resets the data and length variables.
-        bsoncxx::document::view v(data, len);
+        bsoncxx::document::view v(data.get(), len);
         coll.insert_one(v);
-        delete[] data;
-        data = nullptr;
         bytes_read = 0;
         len = 0;
     } else if (bytes_read > len && len >= 4) {
--- a/src/bson_mapper/bson_streambuf.cpp
+++ b/src/bson_mapper/bson_streambuf.cpp
@@ -1,3 +1,17 @@
+// Copyright 2016 MongoDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 #include <cassert>
 #include <cstdio>
 #include <iostream>
--- a/src/bson_mapper/bson_streambuf.cpp
+++ b/src/bson_mapper/bson_streambuf.cpp
@@ -8,46 +8,46 @@
 
 #include "bson_streambuf.hpp"
 
-bson_output_streambuf::bson_output_streambuf(mongocxx::collection *coll){
+bson_output_streambuf::bson_output_streambuf(mongocxx::collection *coll) {
     this->coll = coll;
 }
 
-bson_output_streambuf::~bson_output_streambuf(){
-    if (data != nullptr){
+bson_output_streambuf::~bson_output_streambuf() {
+    if (data != nullptr) {
         delete[] data;
     }
 }
 
-int bson_output_streambuf::underflow(){
+int bson_output_streambuf::underflow() {
     return EOF;
 }
 
-int bson_output_streambuf::overflow(int ch){
+int bson_output_streambuf::overflow(int ch) {
     int result = EOF;
     assert(ch >= 0 && ch <= UCHAR_MAX);
     result = insert(ch);
     return result;
 }
 
-int bson_output_streambuf::insert(int ch){
+int bson_output_streambuf::insert(int ch) {
     bytes_read++;
 
     // for first four bytes, build int32 that contains document size
-    if (bytes_read <= 4){
-        len += (ch << (8 * (bytes_read-1)));
+    if (bytes_read <= 4) {
+        len += (ch << (8 * (bytes_read - 1)));
     }
     // once doc size is received, allocate space.
-    if (bytes_read == 4){
+    if (bytes_read == 4) {
         // TODO perhaps fail unless 0 < len <= MAX_BSON_SIZE?
         data = new uint8_t[len];
         *(size_t *)data = len;
     }
 
-    if (bytes_read > 4){
-        data[bytes_read-1] = (uint8_t)ch;
+    if (bytes_read > 4) {
+        data[bytes_read - 1] = (uint8_t)ch;
     }
 
-    if (bytes_read == len){
+    if (bytes_read == len) {
         // insert document, reset data and length
         bsoncxx::document::view v(data, len);
         coll->insert_one(v);
@@ -55,7 +55,7 @@ int bson_output_streambuf::insert(int ch){
         data = nullptr;
         bytes_read = 0;
         len = 0;
-    } else if (bytes_read > len && len >= 4){
+    } else if (bytes_read > len && len >= 4) {
         return EOF;
     }
     return ch;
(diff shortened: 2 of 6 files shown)


src/bson_mapper/bson_streambuf.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+namespace bson_mapper {
+BSON_MAPPER_INLINE_NAMESPACE_BEGIN
+
+/**
+* A streambuffer that accepts bytes of BSON data,
+* and inserts a document into the given collection when
+* all bytes are sent over.
+*/
+class bson_output_streambuf : public std::streambuf {
```
Machyne: Can/should this be marked `final`?

src/bson_mapper/bson_streambuf.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+    ~bson_output_streambuf();
+
+    /**
+    * This function is called when writing to the stream and the buffer is full.
+    * Since we don't define a buffer, this is called with every character.
+    * @param  ch The byte of BSON to insert.
+    * @return    The inserted byte, or EOF if something failed.
+    */
+    virtual int overflow(int ch);
```
Machyne: Use `override` here instead of `virtual`, that way it ensures that this is overriding a virtual method from the base class.

src/bson_mapper/bson_streambuf.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+    * This function always returns EOF,
+    * since one should not write from an output stream.
+    * @return EOF
+    */
+    virtual int underflow();
```
Machyne: `override`

src/bson_mapper/bson_streambuf.cpp
```diff
@@ -0,0 +1,86 @@
...
+int bson_output_streambuf::insert(int ch) {
+    bytes_read++;
+
+    // for first four bytes, build int32 that contains document size
+    if (bytes_read <= 4) {
+        len += (ch << (8 * (bytes_read - 1)));
+    }
+    // once doc size is received, allocate space.
+    if (bytes_read == 4) {
+        // TODO perhaps fail unless 0 < len <= MAX_BSON_SIZE?
```
Machyne: I agree with this TODO. I think the lower bound should be 4 though, because you've already read 4 bytes.

src/bson_mapper/bson_streambuf.cpp
```diff
@@ -0,0 +1,86 @@
...
+    int result = EOF;
+    assert(ch >= 0 && ch <= UCHAR_MAX);
+    result = insert(ch);
+    return result;
+}
+
+int bson_output_streambuf::insert(int ch) {
+    bytes_read++;
+
+    // for first four bytes, build int32 that contains document size
```
Machyne: Nit: we generally have comments styled as sentences, i.e. start with capitol letter, end with period.
Same throughout.

src/bson_mapper/bson_streambuf.cpp
```diff
@@ -0,0 +1,86 @@
...
+    if (bytes_read <= 4) {
+        len += (ch << (8 * (bytes_read - 1)));
+    }
+    // once doc size is received, allocate space.
+    if (bytes_read == 4) {
+        // TODO perhaps fail unless 0 < len <= MAX_BSON_SIZE?
+        data = new uint8_t[len];
```
Machyne: I think this should be `len-4`.

src/bson_mapper/bson_streambuf.cpp
```diff
@@ -0,0 +1,86 @@
...
+    result = insert(ch);
+    return result;
+}
+
+int bson_output_streambuf::insert(int ch) {
+    bytes_read++;
+
+    // for first four bytes, build int32 that contains document size
+    if (bytes_read <= 4) {
+        len += (ch << (8 * (bytes_read - 1)));
```
Machyne: Need to think about this a bit more, but I'm not immediately convinced that this is safe for non-little-endian.

src/bson_mapper/bson_streambuf.cpp
```diff
@@ -0,0 +1,86 @@
...
+    }
+    // once doc size is received, allocate space.
+    if (bytes_read == 4) {
+        // TODO perhaps fail unless 0 < len <= MAX_BSON_SIZE?
+        data = new uint8_t[len];
+        *(size_t *)data = len;
```
Machyne: Let's do this with a memcpy.

src/bson_mapper/bson_streambuf.cpp
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    if (bytes_read > 4) {
+        data[bytes_read - 1] = (uint8_t)ch;
```
Machyne: We try to avoid c-style casts. Can this be done with a static_cast?
e.g. `static_cast<uint8_t>(ch)`?

src/bson_mapper/bson_streambuf.hpp
```diff
@@ -0,0 +1,78 @@
...
+* all bytes are sent over.
+*/
+class bson_output_streambuf : public std::streambuf {
+   public:
+    /**
+    * Constructs a new BSON Output Streambuffer
+    * that inserts documents into the given collection.
+    * @param coll A pointer to a MongoDB collection.
+    */
+    bson_output_streambuf(mongocxx::collection *coll);
```
Machyne: Who owns `coll` here? Seems like the lifetime of the object has to be managed by the creator of this streambuf. I think we might want to take this by value. `collection` objects seem pretty small, and they have a copy ctor.

src/bson_mapper/bson_streambuf.hpp
```diff
@@ -0,0 +1,78 @@
...
+    * Then, that number of bytes (minus the first 4) are stored in the buffer.
+    * When the data is complete a BSON document view is created, and inserted into
+    * the collection.
+    *
+    * @param  ch The byte to insert.
+    * @return    The byte inserted, or EOF if something failed.
+    */
+    int insert(int ch);
+    mongocxx::collection *coll;
+    uint8_t *data;
```
Machyne: Use `std::unique_ptr<uint8_t>` here, then you can use the default dtor and not worry about calling `delete[]` elsewhere.

src/bson_mapper/bson_streambuf.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include <mongocxx/instance.hpp>
+
+#include "bson_streambuf.hpp"
+
+#include <bson_mapper/config/prelude.hpp>
+
+namespace bson_mapper {
+BSON_MAPPER_INLINE_NAMESPACE_BEGIN
+
+bson_output_streambuf::bson_output_streambuf(mongocxx::collection *coll) {
```
Machyne: You can use a [member initializer list](http://en.cppreference.com/w/cpp/language/initializer_list) here.

src/bson_mapper/bson_streambuf.cpp
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+int bson_output_streambuf::underflow() {
+    return EOF;
+}
+
+int bson_output_streambuf::overflow(int ch) {
+    int result = EOF;
+    assert(ch >= 0 && ch <= UCHAR_MAX);
```
Machyne: It might be more idiomatic here to use the C++ style `std::numeric_limits<uint8_t>::max()`.

Machyne:
Looking good so far! Let's talk about rebasing.

src/bson_mapper/bson_streambuf.hpp
rkargon: As I understand it, `final` would prevent another class from inheriting from `bson_output_streambuf`?

Could there be a situation in which one would want to inherit the class in order to do something fancier at the end of reading the BSON, instead of simply insert_one?

src/bson_mapper/bson_streambuf.hpp
rkargon: done.

src/bson_mapper/bson_streambuf.hpp
rkargon: done.

src/bson_mapper/bson_streambuf.cpp
rkargon: You're right, I tested it with a sample document and used whatever endian-ness it had. Does the BSON spec specify a certain ordering, or can it be detected somehow?

src/bson_mapper/bson_streambuf.cpp
rkargon: done.

src/bson_mapper/bson_streambuf.cpp
rkargon: yup, done.

src/bson_mapper/bson_streambuf.cpp
rkargon: done.

src/bson_mapper/bson_streambuf.cpp
rkargon: done.

src/bson_mapper/bson_streambuf.hpp
rkargon: The class now takes the collection by value, as per the previous comment.

src/bson_mapper/bson_streambuf.hpp
rkargon: Done, the signature is now `bson_output_streambuf(mongocxx::collection coll);`.

src/bson_mapper/bson_streambuf.hpp
Machyne: Fair enough! Sounds good.

Machyne: This class should be marked with `BSON_MAPPER_API`.

src/bson_mapper/bson_streambuf.hpp
```diff
@@ -0,0 +1,76 @@
...
+    * This function inserts a byte of BSON data into the buffer.
+    * The first four bytes are stored in an int, and determine the document size.
+    * Then, that number of bytes (minus the first 4) are stored in the buffer.
+    * When the data is complete a BSON document view is created, and inserted into
+    * the collection.
+    *
+    * @param  ch The byte to insert.
+    * @return    The byte inserted, or EOF if something failed.
+    */
+    int insert(int ch);
```
Machyne: This should be marked `BSON_MAPPER_PRIVATE`.

src/bson_mapper/bson_streambuf.cpp
```diff
@@ -0,0 +1,83 @@
...
+        data.get()[bytes_read - 1] = static_cast<uint8_t>(ch);
+    }
+
+    if (bytes_read == len) {
+        // This inserts the document into the collection, and resets the data and length variables.
+        bsoncxx::document::view v(data.get(), len);
+        coll.insert_one(v);
+        bytes_read = 0;
+        len = 0;
+    } else if (bytes_read > len && len >= 4) {
```
Machyne: What is this else case here for? I'm not sure how we can reach this.

src/bson_mapper/bson_streambuf.hpp
```diff
@@ -0,0 +1,76 @@
...
+* all bytes are sent over.
+*/
+class BSON_MAPPER_API bson_output_streambuf : public std::streambuf {
+   public:
+    /**
+    * Constructs a new BSON Output Streambuffer
+    * that inserts documents into the given collection.
+    * @param coll A pointer to a MongoDB collection.
+    */
+    bson_output_streambuf(mongocxx::collection coll);
```
Machyne: What are your thoughts on taking a callback here rather than a collection? Something like:
```c++
using document_callback = std::function<void(bsoncxx::document::value)>;

bson_output_streambuf(document_callback cb);
```

Thoughts? (cc @adamchel)

src/bson_mapper/bson_streambuf.cpp
rkargon: Good point, I think I added it in while testing earlier and forgot about it.  Now, since `bytes_read` is set to `0` after each document is read, that shouldn't be reachable.

src/bson_mapper/bson_streambuf.hpp
rkargon: I like it, that way one could update the collection instead of insert, or a variety of other actions.

.gitignore
```diff
@@ -9,3 +9,6 @@
 
 # OSX
 .DS_Store
+
+# Editor-specific settings
+.clang_complete
```
rkargon: This is a file with autocomplete settings for my editor.

src/bson_mapper/bson_streambuf.hpp
```diff
@@ -0,0 +1,79 @@
...
+#include <bsoncxx/builder/stream/document.hpp>
+#include <bsoncxx/json.hpp>
+#include <mongocxx/collection.hpp>
+
+#include <bson_mapper/config/prelude.hpp>
+
+namespace bson_mapper {
+BSON_MAPPER_INLINE_NAMESPACE_BEGIN
+
+using document_callback = std::function<void(bsoncxx::document::value)>;
```
Machyne: Move this inside the class

src/bson_mapper/bson_streambuf.cpp
```diff
@@ -0,0 +1,83 @@
...
+        std::memcpy(data.get(), &len, 4);
+    }
+
+    if (bytes_read > 4) {
+        data.get()[bytes_read - 1] = static_cast<uint8_t>(ch);
+    }
+
+    // This creates the document from the given bytes, and calls the user-provided callback.
+    if (bytes_read == len) {
+        bsoncxx::document::value v(std::move(data), len);
```
Machyne: I think lines 70 and 71 could just be `cb({std::move(data), len});`

src/bson_mapper/bson_streambuf.cpp
```diff
@@ -0,0 +1,83 @@
...
+    } else if (bytes_read > len && len >= 4) {
+        return EOF;
+    }
+    return ch;
+}
+
+BSON_MAPPER_INLINE_NAMESPACE_END
+}  // namespace bson_mapper
+
+#include <bson_mapper/config/postlude.hpp>
```
Machyne: Postlude not needed in cpp files

src/bson_mapper/bson_streambuf.hpp
```diff
@@ -0,0 +1,79 @@
...
+    * Then, that number of bytes (minus the first 4) are stored in the buffer.
+    * When the data is complete a BSON document value is created, and passed to the user-provided
+    * callback.
+    *
+    * @param  ch The byte to insert.
+    * @return    The byte inserted, or EOF if something failed.
+    */
+    BSON_MAPPER_PRIVATE int insert(int ch);
+    // This accepts a document::value and returns void.
+    document_callback cb;
```
Machyne: Private member names start with `_` in mongocxx/bsonxxx (I think, please double check), so let's follow that same convention.

Machyne:
Can you add a test for this? :) Looking really good!

src/bson_mapper/bson_streambuf.hpp
```diff
@@ -0,0 +1,79 @@
...
+    */
+    virtual int underflow() override;
+
+   private:
+    /**
+    * This function inserts a byte of BSON data into the buffer.
+    * The first four bytes are stored in an int, and determine the document size.
+    * Then, that number of bytes (minus the first 4) are stored in the buffer.
+    * When the data is complete a BSON document value is created, and passed to the user-provided
+    * callback.
```
adamchel: If I'm understanding the code correctly, the bson_streambuf will be able to accept another document of BSON data after calling the callback because you've reset _len and _bytes_read and you've std::move'd _data. This behavior should probably be documented here. My BSONOutputArchiver can create BSON streams that contain multiple documents so I'd want to know exactly what it's doing with them.

Otherwise, I think it's looking good!

rkargon: You're right, I should add that to the documentation. 
Yeah, you can write multiple documents to the stream one after the other and they'll all be sent to the callback.

.gitignore
```diff
@@ -9,3 +9,15 @@
...
+
+# CMake files
```
Machyne: I think I misunderstood your question from yesterday. If you run cmake from the build directory (i.e. `cd build; cmake ..`), then all of the files should be covered by the .gitignore in the build directory.

src/bson_mapper/bson_streambuf.hpp
```diff
@@ -0,0 +1,86 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <bson_mapper/config/prelude.hpp>
+
+#include <cassert>
```
Machyne: Why did these includes move to the header file? If we don't need them here, I think they are better in the cpp file.

src/bson_mapper/test/bson_streambuf.cpp
```diff
@@ -0,0 +1,63 @@
...
+class doc_validator_callback {
+   public:
+    doc_validator_callback(bsoncxx::document::value doc) : _doc(doc) {
+    }
+
+    /**
+     * This asserts that the given document is the same as the reference document.
+     */
+    void operator()(bsoncxx::document::value v) {
+        REQUIRE(v.view() == _doc.view());
```
Machyne: This looks good! Can you add a counter to this class that increments every time this is called? Then at the end (after line 62), `REQUIRE(cb.count() == 2)`. That way, we can see that this is happening both times.

Machyne:
Two small nits, otherwise LGTM!

src/bson_mapper/test/bson_streambuf.cpp
```diff
@@ -29,17 +32,28 @@ class doc_validator_callback {
...
      * This asserts that the given document is the same as the reference document.
      */
     void operator()(bsoncxx::document::value v) {
+        _count++;
         REQUIRE(v.view() == _doc.view());
     }
 
    private:
+    // Keeps track of the number of documents compared by this validator.
+    static int _count;
```
Machyne: I don't think this should be static.

src/bson_mapper/test/bson_streambuf.cpp
```diff
@@ -29,17 +32,28 @@ class doc_validator_callback {
     doc_validator_callback(bsoncxx::document::value doc) : _doc(doc) {
```
Machyne: add `_count(0)` to the initializer list.

src/bson_mapper/test/bson_streambuf.cpp
```diff
@@ -50,7 +64,7 @@ TEST_CASE("bson_streambuf can faithfully transfer a document",
     int len = bson_view.length();
 
     // set up callback
-    doc_validator_callback cb(bson_obj);
+    auto cb = document_callback(doc_validator_callback(bson_obj));
```
Machyne: Why did this change?

src/bson_mapper/test/bson_streambuf.cpp
```diff
@@ -18,7 +18,10 @@
 #include <bsoncxx/builder/stream/document.hpp>
 #include <bsoncxx/types.hpp>
 
+#include <iostream>
+
 using namespace bson_mapper;
+using document_callback = std::function<void(bsoncxx::document::value)>;
```
Machyne: `using bson_output_streambuf::document_callback;`

src/bson_mapper/test/bson_streambuf.cpp
rkargon: When I made it non-static, I ran into problems with the callback object being copied when passed into the streambuffer.

 I could make bson_streambuf accept a pointer to a document_callback, or have a member function bson_streambuf::callback() that returns such a pointer.

src/bson_mapper/test/bson_streambuf.cpp
Machyne: If you need the `document_callback` here, I think it would be better to just pass in a lambda, e.g.:
```c++
doc_validator_callback validator(bson_obj);
bson_output_streambuf b_buff([&validator](bsoncxx::document::value v) {
    validator.check(v);
});
```

src/bson_mapper/bson_streambuf.hpp
```diff
@@ -61,6 +54,10 @@ class BSON_MAPPER_API bson_output_streambuf : public std::streambuf {
     */
     virtual int underflow() override;
 
+    // document_callback *callback() {
```
Machyne: delete this?

src/bson_mapper/test/bson_streambuf.cpp
```diff
@@ -21,39 +21,36 @@
 #include <iostream>
 
 using namespace bson_mapper;
-using document_callback = std::function<void(bsoncxx::document::value)>;
+using document_callback = bson_output_streambuf::document_callback;
```
Machyne: No longer used?

src/bson_mapper/bson_streambuf.cpp
```diff
@@ -0,0 +1,146 @@
...
+    // or if the given character does not match
+    if (_current == _begin || (ch != EOF && ch != _current[-1])) {
+        return EOF;
+    }
+    _current--;
+    return *_current;
+}
+
+std::streamsize char_array_streambuf::showmanyc() {
+    return _end - _current;
```
Machyne: Could this also just be `return std::distance(_current, _end);`?

src/bson_mapper/bson_streambuf.cpp
```diff
@@ -0,0 +1,146 @@
...
+                break;
+        }
+    }
+    // clamp current pointer to be within the buffer.
+    _current = std::max(std::min(_current, _end), _begin);
+    return std::distance(_begin, _current);
+}
+
+bson_input_streambuf::bson_input_streambuf(const bsoncxx::document::view &v)
+    : char_array_streambuf((const char *)v.data(), v.length()) {
```
Machyne: Does `reinterpret_cast<const char *>(v.data())` work? We try to avoid C style casts.

src/bson_mapper/test/bson_streambuf.cpp
```diff
@@ -0,0 +1,124 @@
...
+    auto bson_view = bson_obj.view();
+    const uint8_t *data = bson_view.data();
+    int len = bson_view.length();
+
+    // set up istream
+    bson_istream bis(bson_view);
+
+    // read bytes of BSON document
+    uint8_t *new_data = new uint8_t[len];
+    bis.read((char *)new_data, len);
```
Machyne: `reinterpret_cast`

src/bson_mapper/bson_streambuf.cpp
```diff
@@ -0,0 +1,146 @@
...
+int char_array_streambuf::uflow() {
+    if (_current == _end) {
+        return EOF;
+    }
+    return *(_current++);
+}
+
+int char_array_streambuf::pbackfail(int ch) {
+    // We can't put back a character if the current pointer is all the way at the beginning,
+    // or if the given character does not match
```
Machyne: End this comment with a period.

src/bson_mapper/bson_streambuf.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+std::streamsize char_array_streambuf::showmanyc() {
+    return std::distance(_current, _end);
+}
+
+std::streampos char_array_streambuf::seekpos(std::streampos sp, std::ios_base::openmode which) {
+    if (which & std::ios_base::in) {
+        _current = std::max(std::min(_begin + sp, _end), _begin);
+    }
+    // clamp current pointer to be within the buffer.
```
Machyne: "Clamp"

Machyne: This comment should probably be up above line 114.

src/bson_mapper/bson_streambuf.cpp
```diff
@@ -0,0 +1,146 @@
...
+    return std::distance(_current, _end);
+}
+
+std::streampos char_array_streambuf::seekoff(std::streamoff off, std::ios_base::seekdir way,
+                                             std::ios_base::openmode which) {
+    // offset is relative to either the beginning, current, or end pointers.
```
Machyne: "Offset"

src/bson_mapper/bson_streambuf.cpp
```diff
@@ -0,0 +1,146 @@
...
+                break;
+            case std::ios_base::cur:
+                _current = _current + off;
+                break;
+            case std::ios_base::end:
+                _current = _end + off;
+                break;
+        }
+    }
+    // clamp current pointer to be within the buffer.
```
Machyne: "Clamp" etc...

src/bson_mapper/test/bson_streambuf.cpp
```diff
@@ -0,0 +1,124 @@
...
+    std::streampos current_pos = is.tellg();
+    is.seekg(10, is.cur);
+    REQUIRE(is.tellg() == (int)current_pos + 10);
+
+    // seeking cannot go past buffer
+    is.seekg(len + 100);
+    REQUIRE(is.tellg() == len);
+}
+
+TEST_CASE("bson_input_streambuf can faithfully send over the bytes of a BSON document",
```
Machyne: Great test!

Machyne:
A few outstanding comment nits, otherwise LGTM!

PR #3, opened 2016-06-10
Preliminary BSON archiver

adamchel:


src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1056 @@
...
+    enum class OutputNodeType { StartObject, InObject, StartArray, InArray };
+
+    typedef bsoncxx::builder::core BSONBuilder;
+
+   public:
+    /**
+    * Construct a BSONOutputArchive that will output to the provided stream.
+    *
+    * @param stream
+    *   The stream to which the archiver will output BSON data to.
```
Machyne: "The stream to which the archiver will output BSON data."

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1056 @@
...
+     *
+     * @param s
+     *    The char* string to save to the archive.
+     */
+    void saveValue(char const* s) {
+        _bsonBuilder.append(s);
+    }
+
+    /**
+     * Write the name of the upcoming element and prepare object/array state
```
Machyne: End with "."

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1056 @@
...
+     * The general workflow of saving to the BSON archive is:
+     *   1. Set the name for the next node to be created, usually done by an NVP.
+     *   2. Start the node.
+     *   3. (if there is data to save) Write the name of the node (this function).
+     *   4. (if there is data to save) Save each element of data (with saveValue).
+     *   5. Finish the node.
+     */
+    void writeName() {
+        if (!_nodeTypeStack.empty()) {
+            OutputNodeType const& nodeType = _nodeTypeStack.top();
```
Machyne: `nodeType` is a bit confusing. Rename to `top` or `topType` or something?

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1056 @@
...
+        _readStream.seekg(0, _readStream.end);
+        size_t streamLength = _readStream.tellg();
+        _readStream.seekg(0, _readStream.beg);
+
+        // Collect every BSON document from the stream.
+        while (_readStream.tellg() < streamLength) {
+            // Determine the size of the BSON document in bytes.
+            // TODO: Only works on little endian.
+            int32_t docsize;
+            char docsize_buf[4];
```
Machyne: `char docsize_buf[sizeof(docsize)];`

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1059 @@
...
+        bool closedObj = false;
+        switch (_nodeTypeStack.top()) {
+            case OutputNodeType::StartArray:
+                _bsonBuilder.open_array();
+            case OutputNodeType::InArray:
+                _bsonBuilder.close_array();
+                closedObj = true;
+                break;
+            case OutputNodeType::StartObject:
+                if (_nodeTypeStack.size() > 1) {
```
Machyne: What conditions lead to this case, where the only element on the stack is `StartObject`?

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1059 @@
...
+
+namespace cereal {
+
+/**
+ * A templated stuct containing a bool value that specifies whether the
+ * provided template parameter is a BSON type.
+ */
+template <class BsonT>
+struct is_bson {
+    static constexpr bool value = std::is_same<BsonT, bsoncxx::types::b_double>::value ||
```
Machyne: Can this be `static constexpr auto`? That might look a bit cleaner.

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1059 @@
...
+     *
+     * @param i
+     *    The int32_t to save to the archive.
+     */
+    void saveValue(std::int32_t i) {
+        _bsonBuilder.append(i);
+    }
+
+    /**
+     * Saves a signed 32 bit int to the current node.
```
Machyne: 64

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1059 @@
...
+     *
+     * @param i
+     *    The int32_t to save to the archive.
```
Machyne: 64

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1059 @@
...
+                if (_nodeTypeStack.size() > 1) {
+                    _bsonBuilder.open_document();
+                }
+            }
+
+            // Elements in arrays do not have names.
+            if (topType == OutputNodeType::InArray) return;
+        }
+
+        if (_nextName == nullptr && _nodeTypeStack.empty() && nodeInRoot) {
```
Machyne: `!_nextName`

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1059 @@
...
+            // This is a document at the root node.
+            // Do nothing since no name is expected.
+            return;
+        } else if (_nextName == nullptr) {
```
Machyne: `!_nextName`

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1059 @@
...
+    /**
+    * Starts a new node, and update the stacks so that
+    * we fetch the correct data when calling search().
+    */
+    void startNode() {
+        if (_nodeTypeStack.empty()) {
+            // If we are in the root node, update the state of the node we're
+            // currently in, but do not do anything else, unless the new node
+            // is loading a BSON array as a root element.
+            if (_nextName != nullptr) {
```
Machyne: `if (_nextName) {`

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1059 @@
...
+     * Loads a BSON datetime from the current node.
+     *
+     * @param val
+     *    The time_point variable into which the datetime will be loaded.
+     */
+    void loadValue(std::chrono::system_clock::time_point& val) {
+        auto bsonVal = search();
+        assert_type(bsonVal, bsoncxx::type::k_date);
+        val = std::chrono::system_clock::time_point(
+            std::chrono::duration_cast<std::chrono::system_clock::duration>(
```
Machyne: Is this duration_cast necessary? Can you just do 
`std::chrono::system_clock::time_point(std::chrono::milliseconds{bsonVal.get_date().value})`?

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1059 @@
...
+    // The current root BSON document being viewed.
+    std::vector<bsoncxx::document::value>::iterator _curBsonDoc;
+
+    // Stack maintaining views of embedded BSON documents.
+    std::stack<bsoncxx::document::view> _embeddedBsonDocStack;
+
+    // Stacks maintaining views of embedded BSON arrays, as well as their
+    // iterators.
+    std::stack<bsoncxx::array::view> _embeddedBsonArrayStack;
+    std::stack<bsoncxx::array::view::iterator> _embeddedBsonArrayIteratorStack;
```
Machyne: I'm expecting a lot of this to change, so holding off on review for now.

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1059 @@
...
+            _nodeTypeStack.pop();
+            ++_curBsonDoc;
+        }
+    }
+
+    /**
+    * Starts a new node, and update the stacks so that
+    * we fetch the correct data when calling search().
+    */
+    void startNode() {
```
Machyne: Assuming this will change, so holding off on review for now.

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1059 @@
...
+}  // namespace cereal
+
+// Register archives for polymorphic support.
+CEREAL_REGISTER_ARCHIVE(cereal::BSONInputArchive)
+CEREAL_REGISTER_ARCHIVE(cereal::BSONOutputArchive)
+
+// Tie input and output archives together.
+CEREAL_SETUP_ARCHIVE_TRAITS(cereal::BSONInputArchive, cereal::BSONOutputArchive)
+
+#endif  // CEREAL_ARCHIVES_BSON_HPP_
```
Machyne: End file with new line.

src/bson_mapper/test/archiver_test.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+    int i;
+
+    std::chrono::system_clock::time_point tp;
+
+    std::ifstream is("data.bson");
+    cereal::BSONInputArchive archive(is);
+
+    REQUIRE_THROWS(archive(CEREAL_NVP(tp), CEREAL_NVP(test_obj_arr), i, a1, a2, b1, c1));
+}
```
Machyne: End file with newline

src/bson_mapper/bson_archiver.hpp
adamchel: This happens when you serialize a class with no members.



For example, the following code:

    struct NoSerializedMembers {
        int32_t a;

        template <class Archive>
        void serialize(Archive& ar) {

        }
    };

    struct DataD {
        NoSerializedMembers m;

        template <class Archive>
        void serialize(Archive& ar) {
            ar(CEREAL_NVP(m));
        }

    };

    TEST_CASE("The BSON archiver successfully serializes a class with no members") {
        std::ofstream os("no_data.bson", std::ios_base::binary);
        cereal::BSONOutputArchive archive(os);

        DataD d;

        archive(d);
    }

produces the following bsondump output:

`{
	"m": {}
}
2016-06-14T11:47:36.308-0400	1 objects found`

This was the behavior of the JSONOutputArchive, so that's why I chose to keep it. Let me know if we should have an exception instead along the lines of "Attempting to serialize empty class."

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/test/archiver_test.cpp
adamchel: Done.

src/bson_mapper/test/archiver_test.cpp
```diff
@@ -0,0 +1,213 @@
...
+        c1.resetValues();
+
+        int i;
+
+        std::chrono::system_clock::time_point tp;
+
+        std::ifstream is("data.bson");
+        cereal::BSONInputArchive archive(is);
+        archive(CEREAL_NVP(tp), CEREAL_NVP(test_obj_arr), CEREAL_NVP(i), a1, a2, b1, c1);
+
```
Machyne: If you implemented an `operator==` for DataA (e.g.), you could make a DataA (a1), set the values, save it, make an empty DataA (a2), load from the save file, then compare a1 and a2. I think that would be a good addition the to the tests you already have.

adamchel: Done.

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1078 @@
...
+                    _embeddedBsonArrayIteratorStack.push(_embeddedBsonArrayStack.top().begin());
+                    _nodeTypeStack.push(InputNodeType::InEmbeddedArray);
+                } else {
+                    throw cereal::Exception(
+                        "Invalid name-value pair in root. Root BSON documents have no key.\n");
+                }
+            } else {
+                _nodeTypeStack.push(InputNodeType::InObject);
+            }
+        } else if (_nodeTypeStack.top() == InputNodeType::InObject ||
```
Machyne: Is there any other choice? If not, just make this `} else {`

adamchel: You're right, done.

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1078 @@
...
+                auto elemFromDoc = _curBsonDoc.view()[nextName];
+
+                if (elemFromDoc) {
+                    return elemFromDoc.get_value();
+                }
+            }
+
+            // If we're in an embedded object, look for the key in the object
+            // at the top of the embedded object stack.
+            if (_nodeTypeStack.top() == InputNodeType::InEmbeddedObject) {
```
Machyne: `else if`

adamchel: Done.

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1078 @@
...
+    /**
+     * Loads a 32 bit int from the current node.
+     *
+     * @param val
+     *    The int32_t variable into which the int will be loaded.
+     */
+    void loadValue(std::int32_t& val) {
+        auto bsonVal = search();
+        assert_type(bsonVal, bsoncxx::type::k_int32);
+        val = bsonVal.get_int32().value;
```
Machyne: It's unfortunate that these types can't all be covered with a template. The line above could be done by instead calling `BsonT::type_id`, but this line I'm not sure. @hanumantmk - any thoughts?

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1078 @@
+#ifndef CEREAL_ARCHIVES_BSON_HPP_
```
hanumantmk: Generally, you'll want to use macros with prefixes specific to your project.  In this case BSON_MAPPER_XXX would do the trick.

Machyne:
This is looking good to me! @hanumantmk could you take a look?

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1078 @@
...
+#include <bsoncxx/types/value.hpp>
+
+namespace cereal {
+/**
+ * A templated stuct containing a bool value that specifies whether the
+ * provided template parameter is a BSON type.
+ */
+template <class BsonT>
+struct is_bson {
+    static constexpr auto value = std::is_same<BsonT, bsoncxx::types::b_double>::value ||
```
hanumantmk: Many of these types don't have a storage (b_utf8 for instance just holds a string_view).

Not sure how that would work (those fields would only be valid if the underlying buffer's bytes hadn't been moved around).

If we want to support having those types in your struct we should probably return a buffer along with them (with the bytes copied out of the stream)

src/bson_mapper/bson_archiver.hpp
adamchel: For the BSON types I was thinking of writing an X-macro along the lines of this:

    #define LOAD_VALUE_FUNC(type) void loadValue(bsoncxx::types::b_##type& val) {
        auto bsonVal = search();
        assert_type(bsonVal, bsoncxx::type::k_##type);
        val = bsonVal.get_##type();
    }

but I was worried it was going to be very unclear. What do you guys think?

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1078 @@
...
+
+    // The name of the next element to be added to the archive.
+    char const* _nextName;
+
+    // A stack maintaining the state of the nodes currently being written.
+    std::stack<OutputNodeType> _nodeTypeStack;
+
+};  // BSONOutputArchive
+
+class BSONInputArchive : public InputArchive<BSONInputArchive>, public traits::TextArchive {
```
hanumantmk: what does inheriting textarchive traits bring us?  the binaryness of bson doesn't make the inheritance obvious to me

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1078 @@
...
+        // Iff we ended up serializing an empty object or array, writeName
+        // will never have been called - so start and then immediately end
+        // the object/array.
+        //
+        // We'll also end any object/arrays we happen to be in.
+
+        bool closedObj = false;
+        switch (_nodeTypeStack.top()) {
+            case OutputNodeType::StartArray:
+                _bsonBuilder.open_array();
```
hanumantmk: I know the comment is supposed to do the job, but a comment that the fall through is intentional would be helpful

src/bson_mapper/bson_archiver.hpp
adamchel: Should I just use #pragma once instead? This was from when I copied over Cereal's JSONArchive

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1078 @@
...
+    /**
+     * Implementations of saveValue that save a BSON type to the current node,
+     * for every BSON type except those that are deprecated and those that are
+     * reserved for internal use.
+     *
+     * @param val
+     *    The BSON typed value to save to the archive.
+     */
+    template <class BsonT, traits::EnableIf<is_bson<BsonT>::value> = traits::sfinae>
+    void saveValue(BsonT val) {
```
hanumantmk: I wonder if you don't want to just have the single overload with anything the core builder will take.

It already has a variety of overloads for append that cover all the overloads that are possible

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1078 @@
...
+     *   5. Finish the node.
+     *
+     * @param nodeInRoot
+     *  Specifies whether this is writing the name of a new node in
+     *  the root level. If this is set to true, no name will be enforced
+     *  since root level documents have no name.
+     */
+    void writeName(bool nodeInRoot = false) {
+        if (!_nodeTypeStack.empty()) {
+            OutputNodeType const& topType = _nodeTypeStack.top();
```
hanumantmk: const auto&

?

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1078 @@
...
+            // This is a document at the root node.
+            // Do nothing since no name is expected.
+            return;
+        } else if (!_nextName) {
+            // Enforce the existence of a name-value pair if this is not a node in root,
+            // or this in an element in root.
+            throw cereal::Exception("Missing a name for current node or element.\n");
+        } else {
+            // Set the key of this element to the name stored by the archiver.
+            _bsonBuilder.key_owned(_nextName);
```
hanumantmk: I wonder what the lifetime issues are for _nextName?

Given that they're known at compile time, I suspect you can use key_view

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1078 @@
...
+    // The name of the next element to be added to the archive.
+    char const* _nextName;
+
+    // A stack maintaining the state of the nodes currently being written.
+    std::stack<OutputNodeType> _nodeTypeStack;
+
+};  // BSONOutputArchive
+
+class BSONInputArchive : public InputArchive<BSONInputArchive>, public traits::TextArchive {
+    typedef std::vector<char> buffer_type;
```
hanumantmk: using buffer_type = std::vector<char>

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1078 @@
...
+        _readStream.read(docsize_buf, sizeof(docsize));
+        std::memcpy(&docsize, docsize_buf, sizeof(docsize));
+
+        // Throw an exception if the end of the stream is prematurely reached.
+        if (_readStream.eof() || !_readStream) {
+            throw cereal::Exception("No more data in BSONInputArchive stream.\n");
+        }
+
+        // Create a buffer to store the BSON document.
+        char* bsonData = new char[docsize];
```
hanumantmk: Prefer:

auto bsonData = bsoncxx::stdx::make_unique<char[]>(docsize);

Or just regular std::make_unique if we want to go to c++14

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1078 @@
...
+
+        // Read the BSON data from the stream into the buffer.
+        _readStream.seekg(-sizeof(docsize), std::ios_base::cur);
```
hanumantmk: rather than relying on rewinding, can't we just copy the 4 bytes we just got into the output buffer and do the read at an offset?

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1078 @@
...
+        _readStream.read(bsonData, docsize);
+
+        // Make sure there were no errors reading the BSON data.
+        if (_readStream.eof() || !_readStream) {
+            throw cereal::Exception("No more data in BSONInputArchive stream.\n");
+        }
+
+        // Store the BSON data of the document in our container of documents.
+        _curBsonDoc =
+            bsoncxx::document::value{reinterpret_cast<uint8_t*>(bsonData),
```
hanumantmk: hmm, if we use the right signature, you could just sink a unique_ptr here

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1078 @@
...
+        }
+
+        // Check to make sure a document has been read.
+        if (!_readFirstDoc) {
+            throw cereal::Exception("Error in element search, never loaded BSON data from file.\n");
+        }
+
+        if (_nextName) {
+            // Reset _nextName
+            std::string nextName = _nextName;
```
hanumantmk: same question about lifetime as before (do we need this copy?)

src/bson_mapper/bson_archiver.hpp
adamchel: Fixed, this was another thing I forgot to take out when modifying the JSON archives.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: I'm actually getting rid of this entirely. I was using vector buffers before this could accept unbounded streams but I've changed this.

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1078 @@
...
+            _nextName = nullptr;
+
+            // If we're in an object in the Root (InObject),
+            // look for the key in the current BSON view.
+            if (_nodeTypeStack.top() == InputNodeType::InObject ||
+                _nodeTypeStack.top() == InputNodeType::InRootElement) {
+                auto elemFromDoc = _curBsonDoc.view()[nextName];
```
hanumantmk: const auto& ?

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1078 @@
...
+
+                if (elemFromDoc) {
+                    return elemFromDoc.get_value();
+                }
+            }
+
+            // If we're in an embedded object, look for the key in the object
+            // at the top of the embedded object stack.
+            if (_nodeTypeStack.top() == InputNodeType::InEmbeddedObject) {
+                auto elemFromDoc = _embeddedBsonDocStack.top()[nextName];
```
hanumantmk: const auto& ?

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1078 @@
...
+                if (elemFromDoc) {
+                    return elemFromDoc.get_value();
+                }
+            }
+
+            // Provide an error message if the key is not found.
+            std::string error_msg;
```
hanumantmk: std::stringstream?

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1078 @@
...
+            error_msg += "No element found with the key ";
+            error_msg += _nextName;
+            error_msg += ".";
+            throw cereal::Exception(error_msg);
+
+        } else if (_nodeTypeStack.top() == InputNodeType::InEmbeddedArray) {
+            // If we're in an array (InEmbeddedArray), retrieve an element from
+            // the array iterator at the top of the stack, and increment it for
+            // the next retrieval.
```
hanumantmk: auto& iter = _embeddedBsonArrayIteratorStack.top();
auto elemFromArr = *iter;
++iter;

is a bit cleaner

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: Done here, and for the other elemFromDoc.

For the elemFromArr I only did "const auto" because if I did const auto&, the reference would point to a different element after incrementing the array iterator.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: key_view seems to work for my test cases, but how can we be positive that it's safe in all cases? Hypothetically, what would happen if a name-value pair's name was set with a std::string's c_str() and the std::string went out of scope before archiving that name-value pair?

adamchel: I feel like there will be cases when doing that is necessary because not every BSON key can be expressed as a C++ variable name.

I'll write a test that tries this.

src/bson_mapper/bson_archiver.hpp
hanumantmk: I'm a fan of the xmacro approach.  It's used for basically the same thing in bsoncxx as it is.

An alternative might be to add templatized getters to bson element.  I.e.

view["field"].get<b_int32>();

That would make this kind of wrapping substantially easier

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1078 @@
...
+     */
+    void setNextName(const char* name) {
+        _nextName = name;
+    }
+
+   private:
+    /**
+     * Throws an exception if the type of v is not the specified type t.
+     */
+    inline void assert_type(bsoncxx::types::value v, bsoncxx::type t) {
```
hanumantmk: const& on value at least?

src/bson_mapper/bson_archiver.hpp
hanumantmk: sure, let's just switch to pragma once

src/bson_mapper/bson_archiver.hpp
hanumantmk: I guess it's a matter of whether or not we need to handle runtime derived names or not.

I'm not terribly in favor of that, but I guess I can imagine a config file that defines the mapping?  Something like that.

Even then, it's reasonable for us to ask that the keys have a longer lifetime than any particular callsite

src/bson_mapper/bson_archiver.hpp
hanumantmk: makes sense

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: I think it would be best if we just used key_view and asked that key strings stay alive until archive() is called for that NVP. Unless we're dealing with something multithreaded, nothing should happen to the pointer between the call to key_view() and append().

Regardless of whether I used key_view or key_owned, the following code:

    int i = 229;

    std::string* i_name = new std::string("hello");

    auto i_nvp = cereal::make_nvp(i_name->c_str(), i);

    delete i_name;
    DataC* garbage = new DataC();


    std::ofstream os("bad_name.bson", std::ios_base::binary);
    cereal::BSONOutputArchive archive(os);

    archive(i_nvp);

produces the following bsondump output:

    {"07f1f77bcf86cd799439011":229}
    2016-06-15T13:13:56.668-0400	1 objects found

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: Got rid of the copy and am just using "const char*" instead of std::string.

src/bson_mapper/bson_archiver.hpp
adamchel: I went with the xmacro for now.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,953 @@
...
+                                  std::is_same<BsonT, bsoncxx::types::b_oid>::value ||
+                                  std::is_same<BsonT, bsoncxx::types::b_bool>::value ||
+                                  std::is_same<BsonT, bsoncxx::types::b_date>::value ||
+                                  std::is_same<BsonT, bsoncxx::types::b_int32>::value ||
+                                  std::is_same<BsonT, bsoncxx::types::b_int64>::value ||
+                                  std::is_same<BsonT, bsoncxx::oid>::value;
+};
+
+/**
+ * A templated stuct containing a bool value that specifies whether the
```
Machyne: "struct"

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,953 @@
...
+        return bsoncxx::document::view{_underlyingBSONData.get(), _underlyingBSONDataSize};
+    }
+
+   private:
+    std::shared_ptr<uint8_t> _underlyingBSONData;
+    size_t _underlyingBSONDataSize;
+};
+
+/**
+ * A templated stuct containing a bool value that specifies whether the
```
Machyne: "struct"

src/bson_mapper/bson_archiver.hpp
adamchel: Added the UnderlyingBSONDataBase class. When a serializable class inherits from this base, deserialization will automatically keep track of the memory associated with the class's bsoncxx view types in a shared_ptr.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: Fixed, but we're now using a bsoncxx::document::view and a shared_ptr.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,958 @@
...
+
+    /**
+     * Starts a new node in the BSON output.
+     *
+     * The node can optionally be given a name by calling setNextName prior to
+     * creating the node. Nodes only need to be started for types that are
+     * objects or arrays.
+     */
+    void startNode() {
+        writeName(true);
```
Machyne: Does this need to do a check here for root? I think always passing true works just because of how `writeName()` works, but it seems to me like this could pass in `true` even when it's not opening an element at the root.

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,958 @@
...
+     */
+    void finishNode() {
+        // Iff we ended up serializing an empty object or array, writeName
+        // will never have been called - therefore the node type at the top
+        // of the stack will be StartArray or StartObject - so start and
+        // then immediately end the documents/array.
+
+        // We'll also end any documents/arrays we happen to be in.
+
+        bool closedObj = false;
```
Machyne: The switch statement covers all four cases, and it seems to me like each case sets this to true. What's the point of this boolean? Is what you actually want a check for `_nodeTypeStack.empty()`?

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,958 @@
...
+        // TODO: Only works on little endian.
+        int32_t docsize;
+        char docsize_buf[sizeof(docsize)];
+        _readStream.read(docsize_buf, sizeof(docsize));
+        std::memcpy(&docsize, docsize_buf, sizeof(docsize));
+
+        // Throw an exception if the end of the stream is prematurely reached.
+        if (_readStream.eof() || !_readStream) {
+            throw cereal::Exception("No more data in BSONInputArchive stream.\n");
+        }
```
Machyne: also throw here if `docsize` < 5

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,991 @@
...
+            // If there is a name, the new node is loading a BSON array as a root element.
+            if (_nextName) {
+                _nodeTypeStack.push(InputNodeType::InRootElement);
+                auto newNode = search();
+
+                if (newNode.type() == bsoncxx::type::k_array) {
+                    _embeddedBsonArrayStack.push(newNode.get_array().value);
+                    _embeddedBsonArrayIteratorStack.push(_embeddedBsonArrayStack.top().begin());
+                    _nodeTypeStack.push(InputNodeType::InEmbeddedArray);
+                } else {
```
Machyne: As per our conversation on this, add an `else if` case to handle `k_document`.

src/bson_mapper/bson_archiver.hpp
adamchel: No, thanks for bringing this up because it made me realize that nodeInRoot is the wrong name for this bool parameter. It should be isNewNode because `writeName()` checks whether it's in root and then checks this variable to decide whether or not to force the existence of a name. I made this change.

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,991 @@
...
+    BSON_LOAD_VALUE_FUNC(utf8)
+    BSON_LOAD_VALUE_FUNC(document)
+    BSON_LOAD_VALUE_FUNC(array)
+    BSON_LOAD_VALUE_FUNC(binary)
+    BSON_LOAD_VALUE_FUNC(oid)
+    BSON_LOAD_VALUE_FUNC(bool)
+    BSON_LOAD_VALUE_FUNC(date)
+    BSON_LOAD_VALUE_FUNC(int32)
+    BSON_LOAD_VALUE_FUNC(int64)
+
```
Machyne: If you keep these x-macros, add an `#undef` afterwards.

src/bson_mapper/bson_archiver.hpp
adamchel: The boolean is no longer necessary, it was an artifact from when I had Root as a NodeType. Theoretically _nodeTypeStack.empty() should never be true but I'll add a check just so we don't segfault.

src/bson_mapper/bson_archiver.hpp
adamchel: Wouldn't an empty BSON document have a docsize of 4?

adamchel: NVM, I see that there is a null byte to finish a document. Done.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1013 @@
...
+/**
+ * A templated struct containing a bool value that specifies whether the
+ * provided template parameter is a BSON type that contains a view.
+ */
+template <class BsonT>
+struct is_bson_view {
+    static constexpr auto value = std::is_same<BsonT, bsoncxx::types::b_utf8>::value ||
+                                  std::is_same<BsonT, bsoncxx::types::b_document>::value ||
+                                  std::is_same<BsonT, bsoncxx::types::b_array>::value ||
+                                  std::is_same<BsonT, bsoncxx::types::b_binary>::value;
```
hanumantmk: regex, dbpointer, code, symbol, codewscope all have views as well

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1013 @@
...
+};
+
+class BSONOutputArchive : public OutputArchive<BSONOutputArchive> {
+    /**
+    * The possible states for the BSON nodes being output by the archive.
+    */
+    enum class OutputNodeType { StartObject, InObject, StartArray, InArray };
+
+    typedef bsoncxx::builder::core BSONBuilder;
```
hanumantmk: using BSONBuilder = bsoncxx::builder::core

is more stylish

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1013 @@
...
+     *    A value t where T is one of the bsoncxx view types.
+     */
+    template <class T>
+    typename std::enable_if<is_bson_view<typename std::decay<T>::type>::value>::type saveValue(
+        T&& t) {
+        if (_curNodeInheritsUnderlyingBSONDataBase.empty() ||
+            !_curNodeInheritsUnderlyingBSONDataBase.top()) {
+            throw cereal::Exception(
+                "Cannot serialize bsoncxx view type (b_utf8, b_document, b_array, b_binary) unless "
+                "that type is wrapped in a class that inherits UnderlyingBSONDataBase.\n");
```
hanumantmk: You don't generally embed a trailing newline in exception strings

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1013 @@
...
+        _underlyingBSONDataSize = size;
+    }
+
+    const bsoncxx::document::view getUnderlyingBSONData() {
+        return bsoncxx::document::view{_underlyingBSONData.get(), _underlyingBSONDataSize};
+    }
+
+   private:
+    std::shared_ptr<uint8_t> _underlyingBSONData;
+    size_t _underlyingBSONDataSize;
```
hanumantmk: Whenever we figure out our endianness story, we can strip out this size_t and rely on the underlying data being wellformed (saves 8 bytes off a struct, which isn't nothing).

Then we can read the size out of the first 4 bytes of the underlying data when we need to make a view

src/bson_mapper/bson_archiver.hpp
adamchel: I specifically chose not to include those types as serializable with the ODM because in bsoncxx/types.hpp, these either have
```
/// @warning
///   This BSON type is used internally by the MongoDB server - use by clients
///   is discouraged.
```
or
```
/// @deprecated
///   A BSON DBPointer (aka DBRef) is still supported but deprecated.
```
I can add them if you think it would be a good idea, but I felt that we would be opening up an unnecessary can of worms if we started supporting these deprecated or internal types.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1013 @@
...
+        _readFirstDoc = true;
+    }
+
+    /**
+     * Searches for the next BSON element to be retrieved and loaded.
+     *
+     * @return the element with _nextName as a key, or if the current node is an
+     *         array, the next element in that array.
+     */
+    inline bsoncxx::types::value search() {
```
hanumantmk: We don't necessarily have to tackle this now, but I wonder how possible it is to invert this loop?

I.e. right now we iterate over the fields we have to populate, looking them up in the bson each time (which involves a linear scan of memory).  Could we invert it so that we fill in fields if they exist as we traverse the bson?  That processing is likely far less expensive for large documents

src/bson_mapper/bson_archiver.hpp
hanumantmk: Technically you have to do more to be really hygenic (pragma push before, then undef before defining yours, then undef, then pragma pop on the way out).

Probably better to disambiguate these (MONGOCXX_ODM_BSON_LOAD_VALUE_FUNC, or whatever other good prefix we're picking for this project) and not worry about it

src/bson_mapper/test/archiver_test.cpp
```diff
@@ -0,0 +1,462 @@
...
+ * Counts the number of keys in a particular document view.
+ *
+ * @param v
+ *      the bsoncxx::document::view to count keys in.
+ *
+ * @return the number of keys in the view.
+ */
+size_t countKeys(bsoncxx::document::view v) {
+    size_t count = 0;
+    for (auto elem = v.begin(); elem != v.end(); ++elem) {
```
hanumantmk: std::distance should do the trick here

src/bson_mapper/test/archiver_test.cpp
```diff
@@ -0,0 +1,462 @@
...
+    int32_t x, y;
+    double z;
+
+    template <class Archive>
+    void serialize(Archive& ar) {
+        ar(CEREAL_NVP(x), CEREAL_NVP(y), CEREAL_NVP(z));
+    }
+
+    bool operator==(const DataA& other) {
+        return x == other.x && y == other.y && z == other.z;
```
hanumantmk: There's a magical trick you can use with std::tuple

return std::tie(x, y, z) == std::tie(other.x, other.y, other.z);

works great when you have to do operator < as well

src/bson_mapper/test/archiver_test.cpp
```diff
@@ -0,0 +1,462 @@
...
+        ar(CEREAL_NVP(test_double), CEREAL_NVP(test_utf8), CEREAL_NVP(test_doc),
+           CEREAL_NVP(test_arr), CEREAL_NVP(test_bin), CEREAL_NVP(test_oid), CEREAL_NVP(test_oid2),
+           CEREAL_NVP(test_bool), CEREAL_NVP(test_date), CEREAL_NVP(test_int32),
+           CEREAL_NVP(test_int64));
+    }
+
+    DataC()
+        : test_double{1.4},
+          test_utf8("cxx-odm"),
+          test_bin{bsoncxx::binary_sub_type::k_binary, sizeof(int),
```
hanumantmk: Not sure if you intend to get different serialization on big/little endian (or if you're comparing this with a fixed point like a binary file).

You might prefer just having some bytes to put in the binary to side skip the issue.  I.e.

uint8_t someBytes[] = { 1, 2, 3, 4 };

src/bson_mapper/test/archiver_test.cpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/test/archiver_test.cpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
adamchel: I'll add a TODO.

src/bson_mapper/test/archiver_test.cpp
adamchel: This was just to get some random binary data. I replaced it with someBytes.

src/bson_mapper/bson_archiver.hpp
adamchel: Per our discussion earlier, I will add support for every BSON type and add the types to the DataC test.

adamchel: Done, and I also made the tests for all the b_ types a bit more comprehensive.

src/bson_mapper/bson_archiver.hpp
adamchel: It seems like Raphael is using MONGO_ODM_, so I'm going to go with that.

src/bson_mapper/bson_archiver.hpp
adamchel: Cereal works by linearly iterating through the fields we have to populate (and recursively iterating through fields in subdocuments), so this would be pretty tricky.

We could theoretically create a `std::map` hash table when we read in the document in `readNextDoc()` and use that in `search()` instead of the `bsoncxx::document::view`'s `[ ]` overload.

I think it would be the most feasible way to do this, because trying to fill in the fields if they exist as we read through the BSON just isn't possible with the way cereal works. What do you think?

hanumantmk: It helps a little bit, but doesn't deal with the fact that we still need to page in the values behind the keys (so we'd save on scanning, but we'd still thrash the cache).

I guess the idea that I have in mind is that as you traverse the fields in the struct (that cereal knows about), you push the requests into a vector.  When you go into a sub doc, you push a "sub doc token".  Then when you leave a doc, you go back in the list of tokens, looping over the bson once, and fill out the instructions by linearly scanning names until you find the open doc token.

I.e.

```cpp
/* Given input data:
{
  foo : 1
  bar : {
    baz : 2
  },
  qux : 3
};
*/

struct Bar {
  int32_t baz;
  template <class Archive>
  void serialize(Archive& ar) {
    ar(CEREAL_NVP(baz)); 
  };
};

struct Data {
  int32_t foo;
  Bar bar;
  int32_t qux;

  template <class Archive>
  void serialize(Archive& ar) {
    ar(CEREAL_NVP(foo), CEREAL_NVP(bar), CEREAL(qux)); 
  };
};

/* I'd expect:
push("foo", types_enum::int32, &foo);
push("bar", types_enum::enter_object);
push("baz", types_enum::int32, &baz);
push(exit_object);
push("qux", types_enum:int32, &qux);

Then when we loop over the bson object, you scan all of the requirements at the correct level (you skip subobject fields) and fill them if you match.  When you hit an enter object, you scan only the records at the correct level until your corresponding exit_object.

As you find records, you erase them from the array.

I suspect this will be faster than a set for numKeys < 1000.
*/
```

adamchel: Filed this as https://jira.mongodb.org/browse/CXX-942.

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1041 @@
...
+#include <string>
+#include <vector>
+
+#include <cereal/cereal.hpp>
+
+#include <bsoncxx/builder/core.hpp>
+#include <bsoncxx/types.hpp>
+#include <bsoncxx/types/value.hpp>
+
+namespace cereal {
```
hanumantmk: I missed this before, but I imagine we probably don't want to put our code in cereal's namespace if we're not going to get upstreamed

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1041 @@
+#pragma once
+
+#include <memory>
```
hanumantmk: include <chrono>

src/bson_mapper/bson_archiver.hpp
adamchel: Yeah, I was thinking that. I added it to the bson_mapper namespace instead.

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/bson_mapper/bson_archiver.hpp
```diff
@@ -0,0 +1,1097 @@
...
+
+    /**
+     * Saves a value of type T to the current node, and throws an exception if we're serializing a
+     * bsoncxx view type without it being wrapped in a class that inherits UnderlyingBSONDataBase.
+     *
+     * @param t
+     *    A value t where T is one of the bsoncxx view types.
+     */
+    template <class T>
+    typename std::enable_if<is_bson_view<typename std::decay<T>::type>::value>::type saveValue(
```
Machyne: What is the `typename std::decay<T>::type` doing here? Can this just be `T`?

adamchel: It has to be there, because it gets rid of any const or reference qualifiers before comparing it to the types in the is_bson_view struct. Without it, the correct overload isn't chosen.

src/bson_mapper/test/archiver_test.cpp
```diff
@@ -0,0 +1,514 @@
...
+    "the BSON archiver enforces the existence of name-value pairs in input archives without "
+    "segfaulting.") {
+    BadDataA a1;
+    DataA a2;
+    DataB b1;
+
+    std::vector<DataA> test_obj_arr;
+
+    DataC c1;
+    c1.resetValues();
```
Machyne: Archiving this after reseting could be dangerous, see this [CXX ticket](https://jira.mongodb.org/browse/CXX-940).

If in the reset for oid's, you switched to the `bsoncxx::oid{bsoncxx::oid::init_tag}` ctor, this problem should go away?

adamchel: Done.

Machyne:
LGTM! 🚢

hanumantmk:
lgtm.

squash it and 🚢

PR #6, opened 2016-06-27
CXXODM-13 aggergate() function wrapper is now templated on return type

rkargon:
This allows people to use the same collection wrapper to do different types of aggregation queries. One can just call `collection_wrapper.aggregate<ResultType>(pipeline, options)`, with whatever `ResultType` they need without creating a new `odm_collection<ResultType>` object.

### DIFF
--- a/src/mongo_odm/odm_collection.hpp
+++ b/src/mongo_odm/odm_collection.hpp
@@ -48,25 +48,28 @@ class odm_collection {
 
     ///
     /// Runs an aggregation framework pipeline against this collection, and returns the results
-    /// as
-    /// de-serialized object.
+    /// as de-serialized objects.
+    /// This function is templated on the result's type, which is not necessarily the same as the
+    /// type of the documents in the colleciton.
     ///
+    /// @tparam Result - The type of the aggregation result.
     /// @param pipeline
     ///   The pipeline of aggregation operations to perform.
     /// @param options
     ///   Optional arguments, see mongocxx::mongocxx::options::aggregate.
     ///
-    /// @return A deserializing_cursor with the results.
+    /// @return A deserializing_cursor<Result> with the results.
     /// @throws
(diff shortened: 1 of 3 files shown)


src/mongo_odm/odm_collection.hpp
```diff
@@ -48,25 +48,28 @@ class odm_collection {
 
     ///
     /// Runs an aggregation framework pipeline against this collection, and returns the results
-    /// as
-    /// de-serialized object.
+    /// as de-serialized object.
```
Machyne: "objects"

src/mongo_odm/odm_collection.hpp
```diff
@@ -48,25 +48,28 @@ class odm_collection {
...
+    /// This function is templated on the result's type, which is not necessarily the same as the
+    /// type of the documents in the colleciton.
     ///
+    /// @tparam Result - The type of the aggregation result.
```
Machyne: Nice use of `@tparam` :)

src/mongo_odm/odm_collection.hpp
```diff
@@ -48,25 +48,28 @@ class odm_collection {
...
-    /// @return A deserializing_cursor with the results.
+    /// @return A deserializing_cursor<Result> with the results.
     /// @throws
     ///   If the operation failed, the returned cursor will throw an mongocxx::exception::query
     ///   when it is iterated.
     ///
     /// @see http://docs.mongodb.org/manual/reference/command/aggregate/
     ///
-    deserializing_cursor<T> aggregate(
+    template <class Result>
```
Machyne: In case someone wants to do an aggregation that returns a cursor of `T`'s, I think we should default this here, e.g.
`template <class Result = T>`

Machyne:
two small nits, but LGTM!

PR #7, opened 2016-06-30
CXXODM-11 Add a CRTP object "model" base class

adamchel:


src/mongo_odm/model.hpp
```diff
@@ -0,0 +1,148 @@
...
+    ///
+    /// @return Cursor with deserialized objects from the collection.
+    /// @throws
+    ///   If the find failed, the returned cursor will throw mongocxx::exception::query when it
+    ///   is iterated.
+    ///
+    /// @see http://docs.mongodb.org/manual/core/read-operations-introduction/
+    ///
+    static deserializing_cursor<T> find(
+        bsoncxx::document::view_or_value&& filter,
```
Machyne: Why not just `bsoncxx::document::view_or_value`?

src/mongo_odm/model.hpp
```diff
@@ -0,0 +1,148 @@
...
+        const mongocxx::options::find& options = mongocxx::options::find()) {
+        return _coll.find(std::forward<bsoncxx::document::view_or_value>(filter),
```
Machyne: This should probably be
`return _coll.find(std::move<bsoncxx::document::view_or_value>(filter), options);`

`std::forward` is useful for universal references, but not here I think.

src/mongo_odm/model.hpp
```diff
@@ -0,0 +1,148 @@
...
+    ///
+    /// @return An optional object that matched the filter.
+    /// @throws mongocxx::exception::query if the operation fails.
+    ///
+    /// @see http://docs.mongodb.org/manual/core/read-operations-introduction/
+    ///
+    static mongocxx::stdx::optional<T> find_one(
+        bsoncxx::document::view_or_value&& filter,
+        const mongocxx::options::find& options = mongocxx::options::find()) {
+        return _coll.find_one(std::forward<bsoncxx::document::view_or_value>(filter),
```
Machyne: ditto

src/mongo_odm/model.hpp
```diff
@@ -0,0 +1,148 @@
...
+    /// @param options
+    ///   Optional arguments, see mongocxx::options::find
+    ///
+    /// @return An optional object that matched the filter.
+    /// @throws mongocxx::exception::query if the operation fails.
+    ///
+    /// @see http://docs.mongodb.org/manual/core/read-operations-introduction/
+    ///
+    static mongocxx::stdx::optional<T> find_one(
+        bsoncxx::document::view_or_value&& filter,
```
Machyne: Again, I think just by value is fine for this type

src/mongo_odm/model.hpp
```diff
@@ -0,0 +1,148 @@
...
+        return _coll.find_one(std::forward<bsoncxx::document::view_or_value>(filter),
+                              std::forward<const mongocxx::options::find>(options));
+    }
+};
+
+template <typename T, typename IdType>
+odm_collection<T> model<T, IdType>::_coll;
+
+MONGO_ODM_INLINE_NAMESPACE_END
+}  // namespace mongo_odm
```
Machyne: newline at end of file

src/mongo_odm/model.hpp
adamchel: Done. I'll read more about rvalues and universal references so I understand this better.

src/mongo_odm/model.hpp
adamchel: Done.

src/mongo_odm/model.hpp
adamchel: Done.

src/mongo_odm/model.hpp
adamchel: Done.

src/mongo_odm/model.hpp
adamchel: Done.

src/mongo_odm/model.hpp
```diff
@@ -0,0 +1,146 @@
...
+#ifdef __APPLE__
+    static odm_collection<T> _coll;
+#else
+    static thread_local odm_collection<T> _coll;
+#endif
+
+   protected:
+    // TODO: if IdType is bsoncxx::oid, the derived class must initialize this until CXX-940 is
+    // resolved.
+    // if IdType is not bsoncxx::oid, the derived class is always responsible for initializing it.
```
Machyne: The hanging indent on these two TODO's is inconsistent. I'm impartial as to whether you want it or not, but please make them match :)

src/mongo_odm/model.hpp
```diff
@@ -0,0 +1,146 @@
...
+    IdType _id;
+
+   public:
+    /**
+     * Returns a copy of the underlying collection.
+     * @return The mongocxx::collection that this object wraps
+     */
+    static mongocxx::collection collection() {
```
Machyne: Should we return this as const ref instead of copy?
Do you think users will want to modify the return of this? I would expect not, but curious to hear your thoughts.

src/mongo_odm/model.hpp
```diff
@@ -0,0 +1,146 @@
...
+        return _coll.collection();
+    }
+
+    /**
+     * Sets the mongocxx::collection mapped to this serializable class.
```
Machyne: This comment is a bit confusing. Consider rewording?
Maybe "Sets the underlying `mongocxx::collection` used to store and load instances of `T`."

src/mongo_odm/model.hpp
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * Performs an update in the database that saves the current T object instance to the
+     * collection mapped to this class.
+     *
+     * In the terms of the CRUD specification, this uses replaceOne with the _id as the sole
+     * argument to the query filter and upsert=true so that objects that aren't already in the
+     * collection are automatically inserted.
+     *
+     * @warning This will fully replace any object in the database with an _id equal to
+     *          this instance's _id.
```
Machyne: Add a link to the [replaceOne](https://docs.mongodb.com/manual/reference/method/db.collection.replaceOne/) docs

src/mongo_odm/model.hpp
```diff
@@ -0,0 +1,146 @@
...
+    ///   Document view representing a document that should match the query.
+    /// @param options
+    ///   Optional arguments, see mongocxx::options::find
+    ///
+    /// @return Cursor with deserialized objects from the collection.
+    /// @throws
+    ///   If the find failed, the returned cursor will throw mongocxx::exception::query when it
+    ///   is iterated.
+    ///
+    /// @see http://docs.mongodb.org/manual/core/read-operations-introduction/
```
Machyne: link to `.com` instead, we recently migrated away from `.org`

src/mongo_odm/model.hpp
```diff
@@ -0,0 +1,146 @@
...
+    ///
+    /// @param filter
+    ///   Document view representing a document that should match the query.
+    /// @param options
+    ///   Optional arguments, see mongocxx::options::find
+    ///
+    /// @return An optional object that matched the filter.
+    /// @throws mongocxx::exception::query if the operation fails.
+    ///
+    /// @see http://docs.mongodb.org/manual/core/read-operations-introduction/
```
Machyne: ditto

src/mongo_odm/model.hpp
```diff
@@ -0,0 +1,146 @@
...
+    ///
+    static deserializing_cursor<T> find(
+        bsoncxx::document::view_or_value filter,
+        const mongocxx::options::find& options = mongocxx::options::find()) {
+        return _coll.find(std::move(filter), options);
+    }
+
+    ///
+    /// Finds a single document in this collection that match the provided filter.
```
Machyne: "matches"

src/mongo_odm/test/model.cpp
```diff
@@ -0,0 +1,108 @@
...
+    a1.save();
+
+    auto query_filter = bsoncxx::builder::stream::document{} << "x" << 16
+                                                             << bsoncxx::builder::stream::finalize;
+
+    auto query_result = DataA::find_one(query_filter.view());
+
+    REQUIRE(query_result);
+
+    query_result.value().remove();
```
Machyne: `query_result->remove();` is more idiomatic

src/mongo_odm/test/model.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+    a1.save();
+
+    auto query_filter = bsoncxx::builder::stream::document{} << "x" << 16
+                                                             << bsoncxx::builder::stream::finalize;
+
+    auto query_result = DataA::find_one(query_filter.view());
+
+    REQUIRE(query_result);
+
```
Machyne: Can you add `REQUIRE(a1 == *query_result);`

src/mongo_odm/test/model.cpp
```diff
@@ -0,0 +1,108 @@
...
+    for (int i = 0; i < 10; ++i) {
+        DataA a1;
+        a1.x = a1.y = a1.z = i;
+        a1.save();
+    }
+
+    auto query_all = bsoncxx::builder::stream::document{} << bsoncxx::builder::stream::finalize;
+
+    // Negate all x values in the collection via pulling them in from a cursor.
+    for (auto a : DataA::find(query_all.view())) {
```
Machyne: `DataA::find({})` should work, then you can delete `query_all`

Machyne: `auto& a` to avoid copies

src/mongo_odm/test/model.cpp
```diff
@@ -0,0 +1,108 @@
...
+        a.x *= -1;
+        a.save();
+    }
+
+    // Sum up the x values in the collection
+    int sum = 0;
+    for (auto a : DataA::find(query_all.view())) {
```
Machyne: ditto

src/mongo_odm/model.hpp
adamchel: Done.

src/mongo_odm/model.hpp
adamchel: Hmm, if this a const ref, then a user can't call any of the underlying CRUD operations or drop(), which I was doing in one of the tests. I think we ideally want users to be able to do everything they need to do directly from the model, so I'm thinking that I'll keep it the way it is for now, but eventually once the model has everything a user would need, we can get rid of this method entirely. What do you think?

src/mongo_odm/model.hpp
adamchel: Done.

src/mongo_odm/model.hpp
adamchel: Done.

src/mongo_odm/model.hpp
adamchel: Done.

src/mongo_odm/model.hpp
adamchel: Done. I actually just changed the link in both places to what the old link was redirecting to.

src/mongo_odm/model.hpp
adamchel: Done.

src/mongo_odm/test/model.cpp

src/mongo_odm/test/model.cpp

src/mongo_odm/test/model.cpp
adamchel: Changed to find({}), but if I change a to auto&, I get the following:

```
/Users/adamchelminski/workspace/mongo-cxx-odm/src/mongo_odm/test/model.cpp:95:16: error: non-const lvalue reference to type 'DataA'
      cannot bind to a temporary of type 'DataA'
    for (auto& a : DataA::find({})) {
               ^ ~
/Users/adamchelminski/workspace/mongo-cxx-odm/src/mongo_odm/deserializing_cursor.hpp:45:14: note: selected 'begin' function with
      iterator type 'mongo_odm::v_noabi::deserializing_cursor<DataA>::iterator'
    iterator begin() {
             ^
```

I'm assuming this means that no copies are being created because the DataA values that the `deserializing_cursor`'s `*` overload returns are non-const non-reference. Based on [RVO](https://en.wikipedia.org/wiki/Return_value_optimization) I would assume that the return value is just being put directly in `a` without copying. Let me know if you think I'm wrong though.

src/mongo_odm/model.hpp
adamchel: I actually changed it to a non-const ref for now.

Machyne: Oops, forgot that those methods weren't const. Returning a copy is good for now.

src/mongo_odm/model.hpp
```diff
@@ -0,0 +1,164 @@
...
+    IdType _id;
+
+   public:
+    /**
+     * Returns a copy of the underlying collection.
+     *
+     * @return A copy of the underlying mongocxx::collection that this class uses to store and
+     * load instances of T.
+     */
+    static const mongocxx::collection& collection() {
```
Machyne: On closer examination and more thought here, I think returning by value (and leaving the caller with a copy) is actually the best thing to do (sorry to make you switch it again). The underlying odm_collection is returning by value, so we could be making a copy here anyway. Collection objects are pretty small, so I think it's best not to worry about it.

src/mongo_odm/test/model.cpp
Machyne: Agreed. Good as is!

Machyne:
One last change for the signature of `collection()`, but otherwise LGTM

PR #8, opened 2016-07-05
Query Builder

rkargon:
A static expression builder for MongoDB queries that allows compile-time type and syntax checking. 

Note: right now this involves a macro to register members and their names, separate from Cereal's `serialize` function, which could seem redundant to users. Perhaps we'd be able to make a macro that does both at once?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,7 +48,7 @@ set(CMAKE_MACOSX_RPATH 1)
 
 # Enforce the C++ standard, and disable extensions
 if(NOT DEFINED CMAKE_CXX_STANDARD)
-    set(CMAKE_CXX_STANDARD 11)
+    set(CMAKE_CXX_STANDARD 14)
 endif()
 
 set(CMAKE_CXX_EXTENSIONS OFF)
--- a/src/mongo_odm/nvp.hpp
+++ b/src/mongo_odm/nvp.hpp
@@ -1,0 +1,207 @@
+// Copyright 2016 MongoDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <mongo_odm/config/prelude.hpp>
+
+#include <cstddef>
+#include <stdexcept>
+#include <tuple>
+#include <type_traits>
+
+#define MONGO_ODM_NVP(x) mongo_odm::makeNvp(&mongo_odm_wrap_base::x, #x)
+
+// Creates serialize() function
+#define MONGO_ODM_SERIALIZE_KEYS                                                                \
+    template <class Archive>                                                                    \
+    void serialize(Archive& ar) {                                                               \
+        mongo_odm_serialize_recur<Archive, 0,                                                   \
+                                  std::tuple_size<decltype(mongo_odm_mapped_fields())>::value>( \
+            ar);                                                                                \
+    }                                                                                           \
+    template <class Archive, size_t I, size_t N>                                                \
+    typename std::enable_if<(I < N), void>::type mongo_odm_serialize_recur(Archive& ar) {       \
+        auto nvp = std::get<I>(mongo_odm_mapped_fields());                                      \
+        ar(cereal::make_nvp(nvp.name, this->*(nvp.t)));                                         \
+        mongo_odm_serialize_recur<Archive, I + 1, N>(ar);                                       \
+    }                                                                                           \
+                                                                                                \
+    template <class Archive, size_t I, size_t N>                                                \
+    typename std::enable_if<(I == N), void>::type mongo_odm_serialize_recur(Archive& ar) {      \
+        ;                                                                                       \
+    }
+
+// Register members and create serialize() function
+#define MONGO_ODM_MAKE_KEYS(Base, ...)                \
+    using mongo_odm_wrap_base = Base;                 \
+    constexpr static auto mongo_odm_mapped_fields() { \
+        return std::make_tuple(__VA_ARGS__);          \
+    }                                                 \
+    MONGO_ODM_SERIALIZE_KEYS
+
+// If using the mongo_odm::model, then also register _id as a field.
+#define MONGO_ODM_MAKE_KEYS_MODEL(Base, ...) \
+    MONGO_ODM_MAKE_KEYS(Base, MONGO_ODM_NVP(_id), __VA_ARGS__)
+
+#define MONGO_ODM_KEY(value) mongo_odm::hasCallIfFieldIsPresent<decltype(&value), &value>::call()
+
+namespace mongo_odm {
+MONGO_ODM_INLINE_NAMESPACE_BEGIN
+
+/**
+ * An object that represents a name-value pair of a member in an object.
+ * It is templated on the class of the member and its type.
+ */
+template <typename Base, typename T>
+struct Nvp {
+    /**
+     * Create a name-value pair from a member pointer and a name.
+     * @param  t    A pointer to the member
+     * @param  name The name of the member
+     */
+    constexpr Nvp(T Base::*t, const char* name) : t(t), name(name) {
+    }
+
+    T Base::*t;
+    const char* name;
+};
+
+/* Create a name-value pair from a object member and its name */
+template <typename Base, typename T>
+Nvp<Base, T> constexpr makeNvp(T Base::*t, const char* name) {
+    return Nvp<Base, T>(t, name);
+}
+
+/**
+ * hasField determines whether a type Base has a member of the given type T as
+ * the Nth member out of M total members which have name value pairs.
+ */
+// By default, if N>=M the index is out of bounds and hasField is false-y.
+template <typename Base, typename T, size_t N, size_t M, bool = (N < M)>
+struct hasField : public std::false_type {};
+
+// Nth member in the Base::fields tuple (i.e. the list of fields for which we
+// have name-value pairs)
+// Must have same type as the given argument.
+template <typename Base, typename T, size_t N, size_t M>
+struct hasField<Base, T, N, M, true>
+    : public std::is_same<T Base::*, decltype(std::get<N>(Base::mongo_odm_mapped_fields()).t)> {};
+
+/**
+ * wrapbool iterates through the fields and returns true if a matching member exists, and false
+ * otherwise.
+ */
+
+// forward declarations for wrapbool
+template <typename Base, typename T, size_t N, size_t M>
+    constexpr std::enable_if_t < N<M && !hasField<Base, T, N, M>::value, bool> wrapbool(T Base::*t);
+
+template <typename Base, typename T, size_t N, size_t M>
+constexpr std::enable_if_t<N == M, bool> wrapbool(T Base::*t);
+
+template <typename Base, typename T, size_t N, size_t M>
+    constexpr std::enable_if_t < N<M && hasField<Base, T, N, M>::value, bool> wrapbool(T Base::*t) {
+    if (std::get<N>(Base::mongo_odm_mapped_fields()).t == t) {
+        return true;
+    } else {
+        return wrapbool<Base, T, N + 1, M>(t);
+    }
+}
+
+template <typename Base, typename T, size_t N, size_t M>
+    constexpr std::enable_if_t <
+    N<M && !hasField<Base, T, N, M>::value, bool> wrapbool(T Base::*t) {
+    return wrapbool<Base, T, N + 1, M>(t);
+}
+
+template <typename Base, typename T, size_t N, size_t M>
+constexpr std::enable_if_t<N == M, bool> wrapbool(T Base::*t) {
+    return false;
+}
+
+/**
+ * wrapimpl uses template arguments N and M to iterate over the fields of a Base
+ * class, and returns the name-value pair corresponding to the given member
+ * field.
+ */
+
+// forward declarations for wrapimpl
+template <typename Base, typename T, size_t N, size_t M>
+constexpr std::enable_if_t<N == M, const Nvp<Base, T>> wrapimpl(T Base::*t);
+
+template <typename Base, typename T, size_t N, size_t M>
+constexpr std::enable_if_t<(N < M) && !hasField<Base, T, N, M>::value, const Nvp<Base, T>> wrapimpl(
+    T Base::*t);
+
+// If Nth field has same type as T, check that it points to the same member.
+// If not, check (N+1)th field.
+template <typename Base, typename T, size_t N, size_t M>
+constexpr std::enable_if_t<(N < M) && hasField<Base, T, N, M>::value, const Nvp<Base, T>> wrapimpl(
+    T Base::*t) {
+    if (std::get<N>(Base::mongo_odm_mapped_fields()).t == t) {
+        return std::get<N>(Base::mongo_odm_mapped_fields());
+    } else {
+        return wrapimpl<Base, T, N + 1, M>(t);
+    }
+}
+
+// If current field doesn't match the type of T, check (N+1)th field.
+template <typename Base, typename T, size_t N, size_t M>
+constexpr std::enable_if_t<(N < M) && !hasField<Base, T, N, M>::value, const Nvp<Base, T>> wrapimpl(
+    T Base::*t) {
+    return wrapimpl<Base, T, N + 1, M>(t);
+}
+
+// If N==M, we've gone past the last field, return nullptr.
+template <typename Base, typename T, size_t N, size_t M>
+constexpr std::enable_if_t<N == M, const Nvp<Base, T>> wrapimpl(T Base::*t) {
+    return Nvp<Base, T>(nullptr, nullptr);
+}
+
+/**
+ * Returns a name-value pair corresponding to the given member pointer.
+ * @tparam Base The class containing the member in question
+ * @tparam T    The type of the member
+ * @param t     A member pointer to an element of type T in the class Base.
+ * @return      The name-value pair corresponding to this member pointer.
+ */
+template <typename Base, typename T>
+constexpr const Nvp<Base, T> wrap(T Base::*t) {
+    return wrapimpl<Base, T, 0, std::tuple_size<decltype(Base::mongo_odm_mapped_fields())>::value>(
+        t);
+}
+
+/**
+ * A struct that has a call() method that returns a name-value pair corresponding
+ * to the given member pointer,
+ * but only if such a member exists.
+ */
+template <typename T, T, typename = void>
+struct hasCallIfFieldIsPresent {};
+
+template <typename Base, typename T, T Base::*ptr>
+struct hasCallIfFieldIsPresent<
+    T Base::*, ptr,
+    std::enable_if_t<wrapbool<
+        Base, T, 0, std::tuple_size<decltype(Base::mongo_odm_mapped_fields())>::value>(ptr)>> {
+    static constexpr const Nvp<Base, T> call() {
+        return wrap(ptr);
+    }
+};
+
+MONGO_ODM_INLINE_NAMESPACE_END
+}  // namespace bson_mapper
+
+#include <mongo_odm/config/postlude.hpp>
--- a/src/mongo_odm/query_builder.hpp
+++ b/src/mongo_odm/query_builder.hpp
@@ -1,0 +1,346 @@
+// Copyright 2016 MongoDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <mongo_odm/config/prelude.hpp>
+
+#include <cstddef>
+#include <stdexcept>
+#include <tuple>
+#include <type_traits>
+
+#include <bsoncxx/builder/core.hpp>
+#include <bsoncxx/view_or_value.hpp>
+
+#include <mongo_odm/nvp.hpp>
+
+namespace mongo_odm {
+MONGO_ODM_INLINE_NAMESPACE_BEGIN
+
+// Forward declarations
+template <typename Base, typename T>
+class NotExpr;
+
+/**
+ * Represents a binary comparison expression between a key and a value. e.g. (User.age > 21).
+ */
+template <typename Base, typename T>
+class ComparisonExpr {
+   public:
+    /**
+     * Constructs a query expression for the given key, value, and comparison type
+     * @param  nvp           A name-value pair corresponding to a key in a document
+     * @param  field         The value that the key is being compared to.
+     * @param  selector_type The type of comparison operator, such at gt (>) or ne (!=).
+     */
+    constexpr ComparisonExpr(const Nvp<Base, T> &nvp, T field, const char *selector_type)
+        : _nvp(nvp), _field(std::move(field)), _selector_type(selector_type) {
+    }
+    /**
+     * Appends this expression to a BSON core builder, as a key-value pair of the form
+     * "key: {$cmp: val}", where $cmp is some comparison operator.
+     * @param builder A BSON core builder
+     */
+    void append_to_bson(bsoncxx::builder::core &builder) const {
+        builder.key_view(_nvp.name);
+        builder.open_document();
+        builder.key_view(_selector_type);
+        builder.append(_field);
+        builder.close_document();
+    }
+
+    /**
+     * Converts the expression to a BSON filter for a query.
+     * The resulting BSON is of the form "{key: {$cmp: val}}".
+     */
+    operator bsoncxx::document::view_or_value() const {
+        auto builder = bsoncxx::builder::core(false);
+        append_to_bson(builder);
+        return builder.extract_document();
+    }
+
+    friend NotExpr<Base, T>;
+
+   private:
+    const Nvp<Base, T> &_nvp;
+    T _field;
+    const char *_selector_type;
+};
+
+/**
+ * This represents an expression with the $not operator, which wraps a comparison expression and
+ * negates it.
+ */
+template <typename Base, typename T>
+class NotExpr {
+   public:
+    /**
+     * Creates a $not expression taht negates the given comparison expression.
+     * @param  expr A comparison expression
+     */
+    constexpr NotExpr(const ComparisonExpr<Base, T> &expr) : _expr(expr) {
+    }
+
+    /**
+     * Appends this expression to a BSON core builder,
+     * as a key-value pair of the form "key: {$not: {$cmp: val}}".
+     * @param builder a BSON core builder
+     */
+    void append_to_bson(bsoncxx::builder::core &builder) const {
+        builder.key_view(_expr._nvp.name);
+        builder.open_document();
+        builder.key_view("$not");
+        builder.open_document();
+        builder.key_view(_expr._selector_type);
+        builder.append(_expr._field);
+        builder.close_document();
+        builder.close_document();
+    }
+
+    /**
+     * Converts the expression to a BSON filter for a query.
+     * The format of the BSON is "{key: {$not: {$cmp: val}}}".
+     */
+    operator bsoncxx::document::view_or_value() const {
+        auto builder = bsoncxx::builder::core(false);
+        append_to_bson(builder);
+        return builder.extract_document();
+    }
+
+   private:
+    const ComparisonExpr<Base, T> _expr;
+};
+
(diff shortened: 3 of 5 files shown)


src/mongo_odm/query_builder.hpp
```diff
@@ -0,0 +1,473 @@
...
+ */
+enum QuerySelector { QS_EQ, QS_GT, QS_GTE, QS_LT, QS_LTE, QS_NE, QS_IN, QS_NIN };
+constexpr char const *const selector_arr[] = {"$eq",  "$gt", "$gte", "$lt",
+                                              "$lte", "$ne", "$in",  "$nin"};
+
+/*
+ * Enumeration of possible logical query operators, as well as a mapping to string values
+ */
+enum LogicalOperator { LO_OR, LO_AND, LO_NOR };
+constexpr char const *const logical_op_arr[] = {"$or", "$and", "$not", "$nor"};
```
Machyne: should `"$not"` be here? Seems like no. It's not in the enum

src/mongo_odm/query_builder.hpp
```diff
@@ -0,0 +1,473 @@
...
+/* Create a name-value pair from a object member and its name */
+template <typename Base, typename T>
+Nvp<Base, T> constexpr makeNvp(T Base::*t, const char *name) {
+    return Nvp<Base, T>(t, name);
+}
+
+/*
+ * Enumeration of possible query selectors, as well as a mapping to string values
+ */
+enum QuerySelector { QS_EQ, QS_GT, QS_GTE, QS_LT, QS_LTE, QS_NE, QS_IN, QS_NIN };
```
Machyne: `enum class` is a little cleaner.

src/mongo_odm/query_builder.hpp
```diff
@@ -0,0 +1,473 @@
...
+ * negates it.
+ */
+template <typename Base, typename T>
+class NotExpr {
+   public:
+    /**
+     * Creates a $not expression taht negates the given comparison expression.
+     * @param  expr A comparison expression
+     */
+    constexpr NotExpr(const ComparisonExpr<Base, T> &expr) : expr(expr) {
```
Machyne: It'd be nice to be able to negate `ExpressionList`s too

Machyne: Never mind, that's not a real thing in mongodb. I always forget that :P

src/mongo_odm/query_builder.hpp
```diff
@@ -0,0 +1,473 @@
...
+     * @param  field         The value that the key is being compared to.
+     * @param  selector_type The type of comparison operator, such at gt (>) or ne (!=).
+     */
+    constexpr ComparisonExpr(const Nvp<Base, T> &nvp, T field, QuerySelector selector_type)
+        : nvp(nvp), field(std::move(field)), selector_type(selector_arr[selector_type]) {
+    }
+
+    const Nvp<Base, T> &nvp;
+    T field;
+    const std::string selector_type;
```
Machyne: Can this be `const char *`? I think the way it is now is making copies.

src/mongo_odm/query_builder.hpp
```diff
@@ -0,0 +1,473 @@
...
+     */
+    operator bsoncxx::document::view_or_value() const {
+        auto builder = bsoncxx::builder::basic::document{};
+        append_to_document_builder(builder);
+        return bsoncxx::document::view_or_value(builder);
+    }
+
+    const Expr1 lhs;
+    const Expr2 rhs;
+    const std::string op;
```
Machyne: ditto `const char *`

src/mongo_odm/query_builder.hpp
rkargon: Haha, I went through that same mental process yesterday.

src/mongo_odm/query_builder.hpp
```diff
@@ -0,0 +1,473 @@
...
+template <typename Base, typename T>
+Nvp<Base, T> constexpr makeNvp(T Base::*t, const char *name) {
+    return Nvp<Base, T>(t, name);
+}
+
+/*
+ * Enumeration of possible query selectors, as well as a mapping to string values
+ */
+enum QuerySelector { QS_EQ, QS_GT, QS_GTE, QS_LT, QS_LTE, QS_NE, QS_IN, QS_NIN };
+constexpr char const *const selector_arr[] = {"$eq",  "$gt", "$gte", "$lt",
```
Machyne: I don't like relying on these to be in the same order. Is there a reason you wanted to use an `enum` at all? If so, maybe write a function that does if/else or construct a `map`.

src/mongo_odm/query_builder.hpp
```diff
@@ -0,0 +1,473 @@
...
+    const std::string op;
+};
+
+/* A templated struct that holds a boolean value.
+* This value is TRUE for types that are query expression,
+* and FALSE for all other types.
+*/
+template <typename>
+struct is_expression {
+    static const bool value = false;
```
Machyne: I think these can be `constexpr`

src/mongo_odm/query_builder.hpp
```diff
@@ -0,0 +1,473 @@
...
+ * @param t     A member pointer to an element of type T in the class Base.
+ * @return      The name-value pair corresponding to this member pointer.
+ */
+template <typename Base, typename T>
+constexpr const Nvp<Base, T> *wrap(T Base::*t) {
+    return wrapimpl<Base, T, 0, std::tuple_size<decltype(Base::fields)>::value>(t);
+}
+
+/*
+A struct that has a call() method that returns a name-value pair corresponding
```
Machyne: add `*` at the start here and below

src/mongo_odm/test/query_builder.cpp
```diff
@@ -0,0 +1,137 @@
...
+    bool y;
+    std::string z;
+    ODM_MAKE_KEYS(Bar, NVP(v), NVP(w), NVP(x1), NVP(x2), NVP(y), NVP(z))
+
+    template <class Archive>
+    void serialize(Archive &ar) {
+        ar(CEREAL_NVP(w), CEREAL_NVP(x1), CEREAL_NVP(x2), CEREAL_NVP(y), CEREAL_NVP(z));
+    }
+};
+ODM_MAKE_KEYS_STORAGE(Bar);
```
Machyne: Doesn't matter yet, but I think this can go away in C++17 with inline variables.

src/mongo_odm/test/query_builder.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+class Bar {
+   public:
+    Bar *v;
+    int64_t w;
+    int x1;
+    int x2;
+    bool y;
+    std::string z;
+    ODM_MAKE_KEYS(Bar, NVP(v), NVP(w), NVP(x1), NVP(x2), NVP(y), NVP(z))
```
Machyne: Can we have `ODM_MAKE_KEYS` also generate the serialize method?

src/mongo_odm/test/query_builder.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include <mongocxx/pipeline.hpp>
+#include <mongocxx/stdx.hpp>
+
+#include <mongo_odm/odm_collection.hpp>
+#include <mongo_odm/query_builder.hpp>
+
+using namespace mongocxx;
+using namespace mongo_odm;
+
+class Bar {
```
Machyne: I think we will be expecting users to inherit from `model`. Can you do that here to make sure it works?

src/mongo_odm/query_builder.hpp
```diff
@@ -0,0 +1,473 @@
...
+#include <bsoncxx/builder/basic/sub_array.hpp>
+#include <bsoncxx/builder/basic/sub_document.hpp>
+#include <bsoncxx/view_or_value.hpp>
+
+#include <bsoncxx/builder/stream/array.hpp>
+
+#define NVP(x) mongo_odm::makeNvp(&wrapBase::x, #x)
+
+#define ODM_MAKE_KEYS(Base, ...) \
+    using wrapBase = Base;       \
```
Machyne: Prefix these names with something like `mongoOdm` just in case users want to use either of these names

src/mongo_odm/test/query_builder.cpp
```diff
@@ -0,0 +1,137 @@
...
+   public:
+    Bar *v;
```
Machyne: What is this used for?

src/mongo_odm/test/query_builder.cpp
```diff
@@ -0,0 +1,137 @@
...
+        REQUIRE(res.value().x1 >= 10);
+
+        res = bar_coll.find_one(!(ODM_KEY(Bar::z) == "hello"));
+        REQUIRE(res);
+        REQUIRE(res.value().z == "goodbye");
+    }
+
+    SECTION("Test expression list.", "[mongo_odm::ExpressionList]") {
+        auto res = bar_coll.find_one(
+            (ODM_KEY(Bar::x1) == 1, ODM_KEY(Bar::x2) == 2, ODM_KEY(Bar::w) >= 444));
```
Machyne: Are the double parens necessary?

src/mongo_odm/query_builder.hpp
```diff
@@ -0,0 +1,473 @@
...
+ * It is templated on the class of the member and its type.
+ */
+template <typename Base, typename T>
+struct Nvp {
+    /**
+     * Create a name-value pair from a member pointer and a name.
+     * @param  t    A pointer to the member
+     * @param  name The name of the member
+     */
+    constexpr Nvp(T Base::*t, const char *name) : t(t), name(name) {
```
adamchel: Totally a nitpicky thing but I think it would be more clear if the pointer-to-member param was `T Base::* t` instead of `T Base::*t` because it makes it more clear what the type of the parameter is and it follows what they do on [cppreference](http://en.cppreference.com/w/cpp/language/pointer#Pointers_to_data_members).

src/mongo_odm/query_builder.hpp
```diff
@@ -0,0 +1,473 @@
...
+    }
+
+    /**
+     * Converts the expression to a BSON filter for a query.
+     * The format of the BSON is "{key: {$not: {$cmp: val}}}".
+     */
+    operator bsoncxx::document::view_or_value() const {
+        auto builder = bsoncxx::builder::basic::document{};
+        append_to_document_builder(builder);
+        return bsoncxx::document::view_or_value(builder);
```
Machyne: `return {builder};` and below

Machyne: This is calling `builder`'s `operator bsoncxx::document::view()`. If builder gets cleaned up after we get a view, this might be an issue. I think you might need to do `builder.extract()`

src/mongo_odm/query_builder.hpp
```diff
@@ -0,0 +1,473 @@
...
+/* Operator overloads for creating and combining expressions */
+
+/* Overload comparison operators for name-value pairs to create expressions.
+ * Separate operators are defined for bool types to prevent confusing implicit casting to bool
+ * for non-bool types.
+ */
+template <typename Base, typename T, typename U,
+          typename = typename std::enable_if_t<!std::is_same<T, bool>::value>>
+constexpr ComparisonExpr<Base, T> operator==(const Nvp<Base, T> &lhs, const U &rhs) {
+    return ComparisonExpr<Base, T>(lhs, rhs, QS_EQ);
```
Machyne: does `return {lhs, rhs, QS_EQ};` work? It looks a little cleaner

src/mongo_odm/test/query_builder.cpp
```diff
@@ -0,0 +1,137 @@
...
+        REQUIRE(res);
+        REQUIRE(res.value().x1 == 1);
+        REQUIRE(res.value().x2 == 2);
+        REQUIRE(res.value().w >= 444);
+    }
+
+    SECTION("Test boolean expressions.", "[mongo_odm::BooleanExpr]") {
+        auto res = bar_coll.find_one(ODM_KEY(Bar::x1) > 9 && ODM_KEY(Bar::x1) < 11);
```
Machyne: This is really cool! 👍

src/mongo_odm/query_builder.hpp
rkargon: The original idea was to avoid passing in invalid operators as strings when constructing the expression objects. But I don't know how much more convenient that would be in the long run, since the operator names are only referred to a handful of times in the code. I can just use string literals instead.

src/mongo_odm/query_builder.hpp
```diff
@@ -0,0 +1,473 @@
...
+    const Nvp<Base, T> &nvp;
+    T field;
+    const std::string selector_type;
+
+    /**
+     * Appends this expression to a BSON document builder, as a key-value pair of the form
+     * "key: {$cmp: val}", where $cmp is some comparison operator.
+     * @param builder A basci BSON document builder
+     */
+    void append_to_document_builder(bsoncxx::builder::basic::sub_document &builder) const {
```
adamchel: Is there any reason this function isn't private in any of the Expr classes?

adamchel: Oh nvm I see that you call it within some of the other classes.

src/mongo_odm/test/query_builder.cpp
rkargon: I was thinking about it, but one would have to map over each argument to the macro and use either `NVP(x)` or `CEREAL_NVP(x)` when creating the member pointer, or in the serialize function, respectively. As far as I know this isn't really possible in the preprocessor. 

@adamchel  Would it be possible to get the BSONArchiver to accept the name-value pairs generated by `NVP(x)` instead of `CERAL_NVP(x)`? 

Obviously it'd be great if users didn't have to register their object members two separate times.

src/mongo_odm/query_builder.hpp
```diff
@@ -0,0 +1,473 @@
...
+    return BooleanExpr<Expr1, Expr2>(lhs, rhs, LO_OR);
+}
+
+/**
+ * hasField determines whether a type Base has a member of the given type T as
+ * the Nth member out of M total members which have name value pairs.
+ */
+// By default, if N>=M the index is out of bounds and hasField is false-y.
+template <typename Base, typename T, size_t N, size_t M,
+          bool = N<M> struct hasField : public std::false_type {};
```
adamchel: for the sake of clarity, you could wrap `N<M` in parentheses so it doesn't look like N templated on M

src/mongo_odm/test/query_builder.cpp
rkargon: Otherwise, the expressions get interpreted as separate arguments to `find_one()`

src/mongo_odm/test/query_builder.cpp
adamchel: @rkargon I don't think I could do that unless we somehow also passed the object instance to the archiver with the NVPs. I don't think your NVPs don't actually contain the values, they just contain relative pointers to the members.

Can we not make a macro along the lines of  `ODM_SETUP(Bar, v, w, x1, x2, y, z)` that would generate the following with the variadic args:

```
    ODM_MAKE_KEYS(Bar, NVP(v), NVP(w), NVP(x1), NVP(x2), NVP(y), NVP(z))

    template <class Archive>
    void serialize(Archive &ar) {
        ar(CEREAL_NVP(w), CEREAL_NVP(x1), CEREAL_NVP(x2), CEREAL_NVP(y), CEREAL_NVP(z));
    }
```

I also want to know what the `Bar *v` member is doing and why it's a key but not serialized.

src/mongo_odm/query_builder.hpp
rkargon: Good point

src/mongo_odm/query_builder.hpp
rkargon: Makes sense to me, but unfortunately clang-format disagrees, I don't know if it's worth the effort of grappling with clang-format every time to keep the whitespace between `*` and `t`. Hope it's not too unclear as is.

src/mongo_odm/test/query_builder.cpp
rkargon: The `Bar *v` is something I forgot to remove earlier from the tests, I'll delete it. 

The C Preprocessor doesn't let you specify individual arguments in an argument list, so we wouldn't be able to do "`NVP(x)` for each `x`". 

However, the `ODM_MAKE_KEYS` macro creates a tuple for the member pointers, it might be possible to write a `serialize` method that iterates over that tuple when called, and use `this->*nvp` to get the value, for each `nvp` in `fields`.

CMakeLists.txt
```diff
@@ -48,7 +48,7 @@ set(CMAKE_MACOSX_RPATH 1)
 
 # Enforce the C++ standard, and disable extensions
 if(NOT DEFINED CMAKE_CXX_STANDARD)
-    set(CMAKE_CXX_STANDARD 11)
+    set(CMAKE_CXX_STANDARD 14)
```
hanumantmk: We should take a look at this eventually to see if we can make this work with C++11 constexpr (for msvc's sake).  I think it's just a matter of inverting a few conditionals to be ternary ops.

src/mongo_odm/query_builder.hpp
```diff
@@ -0,0 +1,466 @@
...
+
+#include <bsoncxx/builder/stream/array.hpp>
+
+#define NVP(x) mongo_odm::makeNvp(&mongo_odm_wrap_base::x, #x)
+
+#define ODM_MAKE_KEYS(Base, ...)      \
+    using mongo_odm_wrap_base = Base; \
+    constexpr static auto fields = std::make_tuple(__VA_ARGS__);
+
+#define ODM_MAKE_KEYS_STORAGE(Base) constexpr decltype(Base::fields) Base::fields
```
hanumantmk: I actually did an update to my initial example that should get rid of this constraint.

https://github.com/hanumantmk/field_member_magic/commit/86c6436754e395be9c1eeaba5cf281c64c62396d

If you make fields a function which returns the tuple, you don't need storage

src/mongo_odm/query_builder.hpp
```diff
@@ -0,0 +1,348 @@
...
+     * Appends this expression to a BSON core builder,
+     * as a key-value pair of the form "key: {$not: {$cmp: val}}".
+     * @param builder a BSON core builder
+     */
+    void append_to_bson(bsoncxx::builder::core &builder) const {
+        builder.key_view(expr.nvp.name);
+        builder.open_document();
+        builder.key_view("$not");
+        builder.open_document();
+        builder.key_view(expr.selector_type);
```
Machyne: Can you change this to use private members and friend classes?

src/mongo_odm/query_builder.hpp
```diff
@@ -0,0 +1,348 @@
...
+    /**
+     * Casts the expression list to a BSON query of  the form { expr1, expr2, ....}
+     */
+    operator bsoncxx::document::view_or_value() const {
+        auto builder = bsoncxx::builder::core(false);
+        append_to_bson(builder);
+        return {builder.extract_document()};
+    }
+
+    const Head head;
```
Machyne: can these be private?

Machyne:
I need to take one more look at the templates, but looking really good!

src/mongo_odm/nvp.hpp
```diff
@@ -0,0 +1,202 @@
...
+#define ODM_SERIALIZE_KEYS                                                           \
+    template <class Archive>                                                         \
+    void serialize(Archive& ar) {                                                    \
+        serialize_recur<Archive, 0, std::tuple_size<decltype(fields())>::value>(ar); \
+    }                                                                                \
+    template <class Archive, size_t I, size_t N>                                     \
+    typename std::enable_if<(I < N), void>::type serialize_recur(Archive& ar) {      \
+        auto nvp = std::get<I>(fields());                                            \
+        ar(cereal::make_nvp(nvp.name, this->*(nvp.t)));                              \
+        serialize_recur<Archive, I + 1, N>(ar);                                      \
```
Machyne: Doing it this way, with a unique call to `ar` per member, I see how loading from BSON works, but won't outputting this to BSON generate a document per member? Or am I wrong about that?

Machyne:
I have a question left about the macro that generates `serialize`, and I want to wait for @hanumantmk to take a look at nvp.hpp, but pending those this looks good!

src/mongo_odm/nvp.hpp
rkargon: Huh. That totally slipped my mind. Unit tests pass, so apparently serialization to BSON works fine. I'm not 100% sure how the serializer works, but does it maintain a stack of current documents, so the BSON isn't finalized until the stack returns to empty? 

 @adamchel ?

adamchel: Calling `ar()` per member should be fine! These invocations of `ar()` are not at the root level so they're not creating a document per member.

src/mongo_odm/nvp.hpp
```diff
@@ -0,0 +1,202 @@
...
+#define NVP(x) mongo_odm::makeNvp(&mongo_odm_wrap_base::x, #x)
+
+// Creates serialize() function
+#define ODM_SERIALIZE_KEYS                                                           \
+    template <class Archive>                                                         \
+    void serialize(Archive& ar) {                                                    \
+        serialize_recur<Archive, 0, std::tuple_size<decltype(fields())>::value>(ar); \
+    }                                                                                \
+    template <class Archive, size_t I, size_t N>                                     \
+    typename std::enable_if<(I < N), void>::type serialize_recur(Archive& ar) {      \
```
Machyne: `mongo_odm_serialize_impl`? Either way, if we're blatting this into the user's objects, let's add the prefix

src/mongo_odm/nvp.hpp
```diff
@@ -0,0 +1,202 @@
...
+        auto nvp = std::get<I>(fields());                                            \
+        ar(cereal::make_nvp(nvp.name, this->*(nvp.t)));                              \
+        serialize_recur<Archive, I + 1, N>(ar);                                      \
+    }                                                                                \
+                                                                                     \
+    template <class Archive, size_t I, size_t N>                                     \
+    typename std::enable_if<(I == N), void>::type serialize_recur(Archive& ar) {     \
```
Machyne: ditto

src/mongo_odm/nvp.hpp
```diff
@@ -0,0 +1,202 @@
...
+        ;                                                                            \
+    }
+
+// Register members and create serialize() function
+#define ODM_MAKE_KEYS(Base, ...)             \
```
Machyne: `MONGO_ODM_MAKE_KEYS`

src/mongo_odm/nvp.hpp
```diff
@@ -0,0 +1,202 @@
...
+    using mongo_odm_wrap_base = Base;        \
+    constexpr static auto fields() {         \
+        return std::make_tuple(__VA_ARGS__); \
+    }                                        \
+    ODM_SERIALIZE_KEYS
+
+// If using the mongo_odm::model, then also register _id as a field.
+#define ODM_MAKE_KEYS_MODEL(Base, ...) ODM_MAKE_KEYS(Base, NVP(_id), __VA_ARGS__)
```
Machyne: `MONGO_...`

src/mongo_odm/nvp.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+#define ODM_KEY(value) mongo_odm::hasCallIfFieldIsPresent<decltype(&value), &value>::call()
```
Machyne: `MONGO_...`?

src/mongo_odm/nvp.hpp
```diff
@@ -0,0 +1,202 @@
...
+                                                                                     \
+    template <class Archive, size_t I, size_t N>                                     \
+    typename std::enable_if<(I == N), void>::type serialize_recur(Archive& ar) {     \
+        ;                                                                            \
+    }
+
+// Register members and create serialize() function
+#define ODM_MAKE_KEYS(Base, ...)             \
+    using mongo_odm_wrap_base = Base;        \
+    constexpr static auto fields() {         \
```
Machyne: `mongo_odm_mapped_fields()`?

src/mongo_odm/nvp.hpp
```diff
@@ -0,0 +1,206 @@
...
+#pragma once
+
+#include <mongo_odm/config/prelude.hpp>
+
+#include <cstddef>
+#include <stdexcept>
+#include <tuple>
+#include <type_traits>
+
+#define NVP(x) mongo_odm::makeNvp(&mongo_odm_wrap_base::x, #x)
```
hanumantmk: name this MONGO_ODM_NVP

src/mongo_odm/query_builder.hpp
```diff
@@ -0,0 +1,356 @@
...
+        builder.key_view(_selector_type);
+        builder.append(_field);
+        builder.close_document();
+    }
+
+    /**
+     * Converts the expression to a BSON filter for a query.
+     * The resulting BSON is of the form "{key: {$cmp: val}}".
+     */
+    operator bsoncxx::document::view_or_value() const {
```
hanumantmk: extract_document gives you a value, just return that

src/mongo_odm/query_builder.hpp
```diff
@@ -0,0 +1,356 @@
...
+     */
+    void append_to_bson(bsoncxx::builder::core &builder) const {
+        _head.append_to_bson(builder);
+        _tail.append_to_bson(builder);
+    }
+
+    /**
+     * Casts the expression list to a BSON query of  the form { expr1, expr2, ....}
+     */
+    operator bsoncxx::document::view_or_value() const {
```
hanumantmk: same idea, just return a value

src/mongo_odm/query_builder.hpp
```diff
@@ -0,0 +1,356 @@
...
+* This value is TRUE for types that are query expression,
+* and FALSE for all other types.
+*/
+template <typename>
+struct is_expression {
+    constexpr static bool value = false;
+};
+
+template <typename Base, typename T>
+struct is_expression<ComparisonExpr<Base, T>> {
```
hanumantmk: you can just inherit from std::true_type, true for all of these

src/mongo_odm/test/query_builder.cpp
```diff
@@ -0,0 +1,166 @@
...
+        REQUIRE(res);
+        REQUIRE(res.value().x1 >= 10);
+
+        res = Bar::find_one(!(MONGO_ODM_KEY(Bar::z) == "hello"));
+        REQUIRE(res);
+        REQUIRE(res.value().z == "goodbye");
+    }
+
+    SECTION("Test expression list.", "[mongo_odm::ExpressionList]") {
+        auto res = Bar::find_one((MONGO_ODM_KEY(Bar::x1) == 1, MONGO_ODM_KEY(Bar::x2) == 2,
```
hanumantmk: I wonder if we shouldn't just skip the expression list version?  it's the same semantics as &&, right?

src/mongo_odm/test/query_builder.cpp
```diff
@@ -0,0 +1,166 @@
...
+                                  MONGO_ODM_KEY(Bar::w) >= 444));
+        REQUIRE(res);
+        REQUIRE(res.value().x1 == 1);
+        REQUIRE(res.value().x2 == 2);
+        REQUIRE(res.value().w >= 444);
+    }
+
+    SECTION("Test boolean expressions.", "[mongo_odm::BooleanExpr]") {
+        auto res = Bar::find_one(MONGO_ODM_KEY(Bar::x1) > 9 && MONGO_ODM_KEY(Bar::x1) < 11);
```
hanumantmk: Maybe tests with multiples && and ||'s mixed and with parens?

src/mongo_odm/nvp.hpp

src/mongo_odm/test/query_builder.cpp
rkargon: && uses the `$and` operator, which apparently sometimes differs from a list of operators:
"MongoDB provides an implicit AND operation when specifying a comma separated list of expressions. Using an explicit AND with the $and operator is necessary when the same field or operator has to be specified in multiple expressions."

src/mongo_odm/query_builder.hpp
rkargon: Since the `mongocxx::collection` methods accept `view_or_value`, I don't think it'd be possible to implicitly convert to expressions to `value`, and then have that be implicitly converted again to `view_or_value`. i.e. The code `col.find_one(ODM_KEY(Foo::x) == 1)` would have to be `col.find_one(view_or_value(ODM_KEY(Foo::x) == 1))`, which I think would be unnecessarily verbose.

At least, when I use `operator bsoncxx::document::value() const`, the tests fail to compile unless I explicitly case expressions to `view_or_value`.

src/mongo_odm/test/query_builder.cpp

src/mongo_odm/query_builder.hpp
```diff
@@ -68,7 +68,7 @@ class ComparisonExpr {
     operator bsoncxx::document::view_or_value() const {
```
hanumantmk: ah, I meant changing the signature to return a bsoncxx::document::value.

Generally we use that type as a way to bind views or values as arguments to a function (we generous in what we accept).  In terms of what we output, we should be strict (we actually always return a value, make that clear from the function signatures).

I.e.

operator bsoncxx::document::value() const {...}

rkargon: I replied to your comment on this from Friday; as far as I understand it, this would require multiple implicit conversions from `ComparisonExpr -> value -> view_or_value` to pass into the collection methods, which the compiler can't do. At any rate, I get syntax errors "no viable conversion from CompairsonExpr to view_or_value..." when I make this change.

hanumantmk:
lgtm % conversation about type trait for IsNvp and sfinae for operators

rkargon:
I'll add the is_nvp type traits stuff to a separate pull request, which includes the nested subfield access code as well, since without nested subfields, I can explicitly specify `Nvp<Base, T>` in the operator functions. So it's not an issue in this pull request.

Machyne:
LGTM

PR #9, opened 2016-07-05
CXXODM-10 Allow stdx::optional types as serializable (and deserializable) fields for the BSON archives

adamchel:


src/bson_mapper/bson_archiver.hpp
```diff
@@ -1024,12 +1109,108 @@ inline void CEREAL_SAVE_FUNCTION_NAME(BSONOutputArchive& ar, cereal::NameValuePa
...
+inline void CEREAL_LOAD_FUNCTION_NAME(BSONInputArchive& ar,
+                                      stdx::optional<bsoncxx::types::b_symbol>& t) {
+    if (ar.willSearchYieldValue()) {
+        bsoncxx::types::b_symbol value{""};
+        ar(value);
+        t.emplace(value);
+    } else {
+        t = stdx::nullopt;
+    }
+}*/
```
rkargon: Were these commented-out lines for the non-default-constructible BSON b_ types? 
Can this be removed  since you have the generalized version that uses `std::aligned_storage`?

adamchel: I talked to Jason about these yesterday. He said that what I'm doing with `std::aligned_storage` is technically undefined behavior so we either have to change the driver to make all b_ types default constructible, or go with an approach similar to what I have in the commented code.

src/bson_mapper/stdx/optional.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include <bsoncxx/stdx/optional.hpp>
+
+namespace bson_mapper {
+namespace stdx {
+
+using bsoncxx::stdx::optional;
+using bsoncxx::stdx::nullopt;
+using bsoncxx::stdx::make_optional;
+}  // namespace stdx
+}  // namespace stdx
```
rkargon: Why are you bringing `stdx::optional` into the bson_mapper namespace?

src/bson_mapper/test/stdx_optional_archiver_test.cpp
```diff
@@ -0,0 +1,299 @@
...
+#include <bsoncxx/json.hpp>
+#include <cereal/types/vector.hpp>
+#include <fstream>
+
+using bson_mapper::stdx::optional;
+using bson_mapper::stdx::make_optional;
+using bson_mapper::stdx::nullopt;
+
+// Optional values.
+struct OptDataA {
```
Machyne: `struct OptDataA : model<OptDataA> {`?

src/bson_mapper/test/stdx_optional_archiver_test.cpp
```diff
@@ -0,0 +1,299 @@
...
+    void serialize(Archive& ar) {
+        ar(CEREAL_NVP(x), CEREAL_NVP(y), CEREAL_NVP(z));
+    }
+
+    bool operator==(const OptDataA& other) const {
+        return std::tie(x, y, z) == std::tie(other.x, other.y, other.z);
+    }
+};
+
+struct SingleElementDataA {
```
Machyne: ditto

src/bson_mapper/test/stdx_optional_archiver_test.cpp
adamchel: I opted not to inherit from the model because the tests in `bson_mapper/test/stdx_optional_archiver_test.cpp` are supposed to be independent from the `mongo_odm` namespace. If you look at `mongo_odm/test/model.cpp`, you'll see that I tested optional fields in a class that inherits from the model.

src/bson_mapper/stdx/optional.hpp
adamchel: They were previously in the `bsoncxx::` namespace which I think would be kind of user unfriendly for a potential odm user.

src/bson_mapper/bson_archiver.hpp
```diff
@@ -512,6 +521,54 @@ class BSONInputArchive : public cereal::InputArchive<BSONInputArchive> {
...
+                    return true;
+                }
+            } else if (_nodeTypeStack.top() == InputNodeType::InEmbeddedObject) {
+                const auto& val = _embeddedBsonDocStack.top()[nextName];
+                if (val) {
+                    _cachedSearchResult = val.get_value();
+                    return true;
+                }
+            }
+        } else if (_nodeTypeStack.top() == InputNodeType::InEmbeddedArray) {
```
Machyne: For these three cases, I think it might be cleaner to:
- check for array first and then throw (for less nesting)
- declare `val`, then assign conditionally on the type stack top
- un-duplicate the caching / return lines

src/mongo_odm/test/model.cpp
```diff
@@ -105,3 +107,59 @@ TEST_CASE(
...
+    b1.z = 1.50;
+
+    b1.save();
+
+    auto query_filter = bsoncxx::builder::stream::document{} << "x" << 16
+                                                             << bsoncxx::builder::stream::finalize;
+
+    auto query_result = DataB::find_one(query_filter.view());
+
+    REQUIRE(query_result);
```
Machyne: Can you add `REQUIRE(!query_result->y)`?

Machyne:
Only small nits. Looking good!

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

src/mongo_odm/test/model.cpp

src/bson_mapper/bson_archiver.hpp
```diff
@@ -542,28 +542,31 @@ class BSONInputArchive : public cereal::InputArchive<BSONInputArchive> {
...
             if (_nodeTypeStack.top() == InputNodeType::InObject ||
                 _nodeTypeStack.top() == InputNodeType::InRootElement) {
-                const auto& val = _curBsonDoc[nextName];
-                if (val) {
-                    _cachedSearchResult = val.get_value();
-                    return true;
-                }
+                val = _curBsonDoc[nextName];
+
             } else if (_nodeTypeStack.top() == InputNodeType::InEmbeddedObject) {
```
Machyne: just `else`?

Machyne:
One small question, but LGTM

src/bson_mapper/bson_archiver.hpp
adamchel: Done.

PR #10, opened 2016-07-06
CXXODM-12 Have cursor wrapper skip documents that don't match schema

rkargon:
The `deserializing_cursor` class now skips documents that don't match the schema (i.e. fail to be deserialized). This makes it easier to use in range-based for loops. 
The iterator also caches objects, so that repeated dereferences don't unnecessarily deserialize the same document over and over.

### DIFF
--- a/src/mongo_odm/deserializing_cursor.hpp
+++ b/src/mongo_odm/deserializing_cursor.hpp
@@ -29,10 +29,9 @@ MONGO_ODM_INLINE_NAMESPACE_BEGIN
 
 /**
  * A class that wraps a mongocxx::cursor. It provides an iterator that deserializes the
- * documents
- * yielded by the underlying mongocxx cursor.
- *
- * TODO what to do if some documents fail deserialization?
+ * documents yielded by the underlying mongocxx cursor.
+ * NOTE: This iterator will skip documents that fail to be deserialized, e.g. due to non-matching
+ * schemas.
  */
 template <class T>
 class deserializing_cursor {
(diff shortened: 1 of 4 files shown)


src/mongo_odm/deserializing_cursor.hpp
```diff
@@ -85,11 +89,38 @@ class deserializing_cursor<T>::iterator : public std::iterator<std::input_iterat
      * underlying collection cursor iterator.
      */
     T operator*() {
-        return bson_mapper::to_obj<T>(*_ci);
+        return _opt ? _opt.value() : bson_mapper::to_obj<T>(*_ci);
```
Machyne: How can `_opt` be unset here? It seems like either we will have a cached value or the cursor will be empty.

src/mongo_odm/test/deserializing_cursor.cpp
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    template <class Archive>
+    void serialize(Archive& ar) {
+        ar(CEREAL_NVP(a), CEREAL_NVP(b), CEREAL_NVP(c));
+    }
+};
+
+// Represents an aggregation result
+class FooResult {
```
Machyne: unused in this test?

rkargon: whoops, this was copied in from another test file by mistake.

src/mongo_odm/deserializing_cursor.hpp
rkargon: Good point, replaced with `return _opt.value()`.

Machyne:
LGTM!
Let's link a new ticket to revisit this when we implement more complex document validation

PR #11, opened 2016-07-07
CXXODM-35 Perform save() in model with updateOne and $set instead of replaceOne

adamchel:


src/bson_mapper/bson_archiver.hpp
```diff
@@ -220,20 +234,37 @@ class BSONOutputArchive : public cereal::OutputArchive<BSONOutputArchive> {
...
+                // array somewhere.
+                if (!_dotNotationMode || _arrayNestingLevel > 0) {
+                    if (_nodeTypeStack.size() > 1) {
+                        _bsonBuilder.close_document();
+                    } else if (_objAsRootElement) {
+                        _bsonBuilder.close_document();
+                    }
+                } else if (_nodeTypeStack.size() > 1) {
+                    // Pop the name of this embedded document off the stack.
+                    _embeddedNameStack.pop_back();
```
Machyne: Will this always be safe? for example in objects nested in arrays?

src/mongo_odm/test/model.cpp
```diff
@@ -164,3 +157,210 @@ TEST_CASE(
...
+
+    e_with_embedded_y.save();
+
+    query_filter = bsoncxx::builder::stream::document{} << "_id" << e_with_embedded_y.getID()
+                                                        << bsoncxx::builder::stream::finalize;
+
+    query_result = DataE::find_one(query_filter.view());
+
+    REQUIRE(query_result);
+    REQUIRE(query_result->m.x);
```
Machyne: Can you add
```c++
REQUIRE(e_with_embedded_x.m.x == *query_result->m.x);
REQUIRE(e_with_embedded_y.m.y == *query_result->m.y);
```

src/bson_mapper/bson_archiver.hpp
adamchel: I'm pretty confident it will be safe, as it's only executed when `dotNotationMode` is true and when `_arrayNestingLevel == 0`, though I could add a check for `_embeddedNameStack.empty()` before `_embeddedNameStack.pop_back()` to avoid a segfault.

The case you mention is actually tested in one of the test cases ("the BSON archiver successfully serializes embedded classes with dot notation"), which includes an array of objects alongside regular embedded objects. If we're in an object nested in an array, `_arrayNestingLevel` will be greater than 0 and the first `if` is invoked.

Machyne: Ah, it's caught on line 259. Great, no need to add more checks here – that's what I had missed.

src/mongo_odm/model.hpp
```diff
@@ -24,8 +24,51 @@
...
+
+/**
+ * Specialization of mongo_odm::model base class for IdType=bsoncxx::oid that provides a default
+ * constructor for _id which safely constructs a bsoncxx::oid.
+ */
+template <>
+class model_odm_base<bsoncxx::oid> {
+   public:
+    template <typename... Ts, typename = std::enable_if_t<!Widget<model_odm_base, Ts...>::value>>
+    model_odm_base(Ts&&... ts) : _id(std::forward<Ts>(ts)...) {
```
Machyne: @hanumantmk PTAL at this ctor / inheritance templating :)
looks good to me though

src/mongo_odm/model.hpp
```diff
@@ -71,6 +112,12 @@ class model {
      *
      * @warning This must be called with a new mongocxx::collection instance for every thread for
      *          the ODM to be thread-safe.
+     *
+     * @warning The mongocxx::client collection with which the passed-in mongocxx::collection was
```
Machyne: Maybe:
"The parent mongocxx::client from which the mongocxx::collection argument was created must outlive any of this model's CRUD methods. If the client object goes out of scope, a new collection must be passed to this method before using any CRUD methods."

adamchel: Done.

src/mongo_odm/model.hpp
```diff
@@ -24,8 +24,51 @@
...
 MONGO_ODM_INLINE_NAMESPACE_BEGIN
 
+/**
+ * Helper type trait widget that helps properly forward arguments to _id constructor.
+ */
+template <typename... Ts>
+struct Widget : public std::false_type {};
+
+template <typename T, typename T2, typename... Ts>
+struct Widget<T, T2, Ts...> : public std::is_same<T, std::decay_t<T2>> {};
```
hanumantmk: Let's name this FirstTypeIsTheSame.  Or pick a name, but don't leave it called Widget

src/mongo_odm/model.hpp
```diff
@@ -24,8 +24,51 @@
...
+
+/**
+ * Base class for mongo_odm::model that provides _id and _id construction.
+ * TODO: Once CXX-940 is resolved, this model_odm_base will no longer be necessary and the argument
+ *       forwarding logic can take place in the mongo_odm::model class.
+ */
+template <typename IdType>
+class model_odm_base {
+   public:
+    template <typename... Ts, typename = std::enable_if_t<!Widget<model_odm_base, Ts...>::value>>
```
hanumantmk: Document why we need this

src/mongo_odm/model.hpp
adamchel: Done.

Machyne: Did you mean to push something here?

src/mongo_odm/model.hpp
adamchel: Done.

src/mongo_odm/model.hpp
adamchel: Yep, pushing it now.

Machyne:
LGTM!

hanumantmk:
lgtm

adamchel:
Merged as https://github.com/mongodb/mongo-cxx-odm/commit/8f6ac231cff96a04f21ba154f0bb7b4c6a551ba4

PR #12, opened 2016-07-13
CXXODM-39 Set up a Travis CI Integration and get it passing on non-GCC variants

adamchel:


.travis.yml
```diff
@@ -0,0 +1,104 @@
...
+
+    # Mongo C Driver
+    - curl -O -L https://github.com/mongodb/mongo-c-driver/releases/download/${CDRIVER_VERSION}/mongo-c-driver-${CDRIVER_VERSION}.tar.gz
+
+    # Mongo C++ Driver
+    - git clone -b 'master' --single-branch https://github.com/mongodb/mongo-cxx-driver.git
+
+install:
+    # GCC
+    - if [ "$CXX" = "g++" ]; then sudo apt-get install -qq g++-6; export CXX="g++-6"; export CC="gcc-6"; export NO_ERROR_ON_MAYBE_UNINITIALIZED="-Wno-error=maybe-uninitialized"; fi
```
Machyne: if you add something like
`if [ "$CONFIG" == "Debug" ]; then export NO_ERROR_ON_MAYBE_UNINITIALIZED=""; fi`
then you can eliminate some duplication below

.travis.yml
```diff
@@ -0,0 +1,104 @@
...
+
+before_script:
+    - $CC --version
+    - $CXX --version
+
+    # Build the odm and the tests
+    - cd build
+    - if [ "$CONFIG" == "Release" ]; then ${CMAKE_BINARY} -DCMAKE_BUILD_TYPE=$CONFIG -DCMAKE_MODULE_PATH=/home/travis/build/mongodb/mongo-cxx-odm/mongo-cxx-driver/cmake -DCMAKE_INSTALL_PREFIX=/home/travis/install -DCMAKE_C_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers ${NO_ERROR_ON_MAYBE_UNINITIALIZED}" -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers ${NO_ERROR_ON_MAYBE_UNINITIALIZED}" ..; fi
+    - if [ "$CONFIG" == "Debug" ]; then ${CMAKE_BINARY} -DCMAKE_BUILD_TYPE=$CONFIG -DCMAKE_MODULE_PATH=/home/travis/build/mongodb/mongo-cxx-odm/mongo-cxx-driver/cmake -DCMAKE_INSTALL_PREFIX=/home/travis/install -DCMAKE_C_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers" -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers" ..; fi
```
Machyne: I think this can go away if you set `NO_ERROR_ON_MAYBE_UNINITIALIZED` to empty string above

.travis.yml
```diff
@@ -0,0 +1,104 @@
...
+    - make all
+
+    # Run bson_mapper tests with catch
+    - ./src/bson_mapper/test/test_bson_mapper
+
+    # Run mongo_odm tests with catch
+    - ./src/mongo_odm/test/test_mongo_odm
+
+    # Install headers and libs for the examples
+    - sudo make install
```
Machyne: I would like to figure out how to do this without root. Our users should be able to install without sudo as long as they configured the directory properly, so let's get that working here too to be sure.

.travis.yml
adamchel: Done.

src/bson_mapper/bson_streambuf.hpp
```diff
@@ -111,12 +111,12 @@ class BSON_MAPPER_API char_array_streambuf : public std::streambuf {
      * This reads and returns a character from the buffer, and increments the read pointer.
      * @return [description]
      */
-    virtual BSON_MAPPER_PRIVATE int underflow() override;
+    int underflow() final;
```
Machyne: Should these all be `final override`? That way they must override a virtual function in the base class.

adamchel: Done.

Machyne:
@adamchel - if we need to fix the sudo issues when we switch to containers anyway, then my last comment could be addressed there. If you want to go that route, then this LGTM!

adamchel:
Okay, sounds good!

PR #14, opened 2016-07-19
(CXXODM-17, CXXODM-40, CXXODM-41) Update builder

rkargon:
Added update operators to the query builder, as well as query operators that don't translate to C++ operators (e.g.`$nin`). 
This pull request also includes CXXODM-40 (accessing child fields) and CXXODM-41 (support for optional types)


src/mongo_odm/nvp.hpp
```diff
@@ -21,6 +21,24 @@
...
 #include <type_traits>
 
+#include <bsoncxx/types.hpp>
+
+#include <mongo_odm/util.hpp>
+
+// Preprocessor templates for manipulating multiple arguments.
+#define PP_NARG(...) PP_NARG_(__VA_ARGS__, PP_RSEQ_N())
+#define PP_NARG_(...) PP_ARG_N(__VA_ARGS__)
+#define PP_ARG_N(_1, _2, _3, _4, _5, _6, _7, _8, _9, _10, _11, _12, _13, _14, _15, _16, _17, _18, \
```
Machyne: Max depth here only needs to be 52 I think-
* 50 for mongodb bson depth
* 1 for classname
* 1 for count

Is that right?

src/mongo_odm/nvp.hpp
```diff
@@ -21,6 +21,24 @@
 #include <tuple>
 #include <type_traits>
 
+#include <bsoncxx/types.hpp>
+
+#include <mongo_odm/util.hpp>
+
+// Preprocessor templates for manipulating multiple arguments.
+#define PP_NARG(...) PP_NARG_(__VA_ARGS__, PP_RSEQ_N())
```
Machyne: For all these `PP_*` macros, are they being used outside of this file? if not, surround this with push macro / undef / pop macro so we don't expose these. (see the libmongocxx prelude/postlude for good examples of this)

src/mongo_odm/nvp.hpp
```diff
@@ -56,16 +74,83 @@
...
+
+#define MONGO_ODM_CHILD1(base, field1) MONGO_ODM_KEY_BY_VALUE(&base::field1)
+
+#define MONGO_ODM_CHILD2(base, field1, field2)                                               \
+    makeNvpWithParent(                                                                       \
+        MONGO_ODM_KEY_BY_VALUE(                                                              \
+            &std::decay_t<typename decltype(MONGO_ODM_CHILD1(base, field1))::type>::field2), \
+        MONGO_ODM_CHILD1(base, field1))
+
+#define MONGO_ODM_CHILD3(base, field1, field2, field3)                                         \
```
Machyne: As much as I hate to say this, I think you probably should go all the way to 50 here for completeness. Maybe move these to a separate header, write a script to generate them, and then never touch it again?

src/mongo_odm/nvp.hpp
rkargon: Ah, it didn't occur to me to take into account the maximum depth of BSON. How about I limit the depth of `MONGO_ODM_CHILD*`, and keep `PP_ARG` as is, in case it gets used later?

src/mongo_odm/nvp.hpp
hanumantmk: Actually, let's drag them out into their own file and namespace the macros.

Nothing about the implementation requires them to be in this particular header

src/mongo_odm/nvp.hpp
hanumantmk: Actually, you may want to support deeper.

Specifically: https://docs.mongodb.com/manual/reference/limits/#Nested-Depth-for-BSON-Documents
"MongoDB supports no more than 100 levels of nesting for BSON documents."

src/mongo_odm/nvp.hpp
hanumantmk: I'm in agreement.  Do 100 levels check in the generated header and also the check in the script that produces the header.  Probably worth adding a target for the header generation as well in cmake

src/mongo_odm/nvp.hpp
```diff
@@ -74,17 +159,593 @@ struct Nvp {
...
     }
 
+    /**
+     * Assignment operator that creaptes an update expression to set the field to a given value.
+     */
+    constexpr UpdateExpr<Nvp<Base, T>, no_opt_type> operator=(const no_opt_type& val) const {
+        return {*this, val, "$set"};
+    }
+
+    virtual std::string get_name() const {
```
hanumantmk: Why force the std::string creation?  Can we use the bsoncxx::stdx::string_view, or if we want to avoid that, just stay in const char* land?

hanumantmk: Also, should this actually be virtual? Or is this just for the final concrete type and crtp magic in NvpCRTP?

src/mongo_odm/nvp.hpp
```diff
@@ -74,17 +159,593 @@ struct Nvp {
...
+    constexpr UpdateExpr<NvpChild<Base, T, Parent>, no_opt_type> operator=(
+        const no_opt_type& val) const {
+        return {*this, val, "$set"};
+    }
+
+    /**
+     * Returns the qualified name of this field in dot notation, i.e. "parent.child".
+     * @return A string containing the name of this field in dot notation.
+     */
+    std::string get_name() const {
```
hanumantmk: ah, I see why you went with std::string.

You're going to be a lot better off providing something less heavy weight though.  Maybe an iterator that yields const char*'s?  That, or if you want something simple, just an ostream overload.  That way you can build the stringstream builder at the top and save yourself DEPTH copies

src/mongo_odm/nvp.hpp
```diff
@@ -74,17 +159,593 @@ struct Nvp {
...
+        std::string full_name;
+        full_name += (parent.get_name() + ".");
+        full_name += this->name;
+        return full_name;
+    }
+
+    T Base::*t;
+    const char* name;
+    const Parent parent;
```
hanumantmk: How well does this work if you capture parent by const& instead of value?

src/mongo_odm/nvp.hpp
```diff
@@ -74,17 +159,593 @@ struct Nvp {
...
+ * These functions are identical between Nvp<...> and NvpChild<...>, but their return types are
+ * templated on the Nvp's types, so they are defined here using CRTP.
+ * @tparam NvpT The type of the name-value pair
+ * @tparam T    The type of the field referred to by the name-value pair.
+ */
+template <typename NvpT, typename T>
+class NvpCRTP {
+   public:
+    // In case this field is wrapped in an optional, store the underlying type.
+    using no_opt_type = remove_optional_t<T>;
```
hanumantmk: The other type's all have a type = T alias.  Maybe here too for consistency?

src/mongo_odm/nvp.hpp
```diff
@@ -74,17 +159,593 @@ struct Nvp {
...
+
+    // Type trait that checks if the given iterable class contains the same value type as either
+    // a) this Nvp's type, or
+    // b) this Nvp's value type, if this Nvp is also an iterable.
+    template <typename Iterable, typename Default = void>
+    using enable_if_matching_iterable_t = typename std::enable_if<
```
hanumantmk: Make this trait private in the class.  No need to leak the symbol

src/mongo_odm/nvp.hpp
```diff
@@ -74,17 +159,593 @@ struct Nvp {
...
+ * @tparam T    The type of the field.
+ */
+template <typename T>
+class FreeNvp : public NvpCRTP<FreeNvp<T>, T> {
+   public:
+    // In case this corresponds to an optional type, store the underlying type.
+    using no_opt_type = remove_optional_t<T>;
+    using type = T;
+
+    constexpr UpdateExpr<FreeNvp<T>, no_opt_type> operator=(const no_opt_type& val) const {
```
Machyne: Do we ever need this for updates?

src/mongo_odm/nvp.hpp
```diff
@@ -74,17 +159,593 @@ struct Nvp {
...
+    }
+
+    /**
+     * Creates a sort expression that sorts documents by this field.
+     * @param ascending Whether to sort by ascending order (+1 in MongoDB syntax).
+     *                  If false, sorts by descending order (-1 in MongoDB syntax).
+     * @return a SortExpr that reprsents the sort expression {field: +/-1}.
+     */
+    constexpr SortExpr<NvpT> sort(bool ascending) const {
+        return {*static_cast<const NvpT*>(this), ascending};
```
Machyne: Maybe add a member pointer:
`const NvpT* self;`
And then initialize at construction to
`static_cast<const NvpT*>(this)`

src/mongo_odm/nvp.hpp
```diff
@@ -74,17 +159,593 @@ struct Nvp {
...
+    constexpr UpdateExpr<NvpT, no_opt_type> max(const no_opt_type& val) const {
+        return {*static_cast<const NvpT*>(this), val, "$max"};
+    }
+
+    /**
+     * Creats an expression that sets a date value to the current date.
+     * This is only enabled for std::chrono::time/duration values, as well as b_date.
+     */
+    template <typename U = no_opt_type>
+    constexpr typename std::enable_if<is_date<U>::value, CurrentDateExpr<NvpT>>::type current_date()
```
Machyne: Instead of making this a member function, can this be an overload of `operator=()` for type `odm_current_date_t` and then also provide a `const odm_current_date_t odm_current_date;`

src/mongo_odm/nvp.hpp
```diff
@@ -74,17 +159,593 @@ struct Nvp {
...
+        const {
+        return {*static_cast<const NvpT*>(this), true};
+    }
+
+    /**
+     * Creats an expression that sets a b_timestamp value to the current date.
+     * This is only enabled for b_timestamp values. (Use of which is discouraged by users.)
+     */
+    template <typename U = no_opt_type>
+    constexpr typename std::enable_if<std::is_same<bsoncxx::types::b_timestamp, U>::value,
```
Machyne: ditto, but provide a different overload based off of underlying type

src/mongo_odm/nvp.hpp
```diff
@@ -56,16 +74,83 @@
...
+    makeNvpWithParent(MONGO_ODM_KEY_BY_VALUE(&std::decay_t<typename decltype(MONGO_ODM_CHILD2( \
+                                                 base, field1, field2))::type>::field3),       \
+                      MONGO_ODM_CHILD2(base, field1, field2))
+
+#define PASTE_IMPL(s1, s2) s1##s2
+#define PASTE(s1, s2) PASTE_IMPL(s1, s2)
+
+#define MONGO_ODM_CHILD(type, ...) PASTE(MONGO_ODM_CHILD, PP_NARG(__VA_ARGS__))(type, __VA_ARGS__)
 
 namespace mongo_odm {
```
hanumantmk: I think it's going to be a good idea to cut out an impl namespace inside of mongo_odm and move all of the utility stuff we have kicking around into that.  Consider moving any types/functions it would never make sense for a user to directly name into that namespace.

src/mongo_odm/query_builder.hpp
```diff
@@ -25,93 +26,453 @@
...
+struct is_query_expression;
+template <typename>
+struct is_update_expression;
+
+/**
+ * Type trait that contains true if a type T can be appended to a BSON builder using
+ * builder.append(T val);
+ */
+template <typename T>
+constexpr auto is_bson_appendable_impl(int)
```
Machyne: Does this need to take an argument?

hanumantmk: Very nice :)

src/mongo_odm/query_builder.hpp
```diff
@@ -25,93 +26,453 @@
...
+// Forward declarations for Expression type trait structs
+template <typename>
+struct is_sort_expression;
+template <typename>
+struct is_query_expression;
+template <typename>
+struct is_update_expression;
+
+/**
+ * Type trait that contains true if a type T can be appended to a BSON builder using
```
Machyne: It would be nice to consolidate type traits when possible

src/mongo_odm/query_builder.hpp
```diff
@@ -25,93 +26,453 @@
...
+    return {};
+}
+
+template <typename T>
+constexpr std::false_type is_bson_appendable_impl(...) {
+    return {};
+}
+
+template <typename T>
+using is_bson_appendable = decltype(is_bson_appendable_impl<T>(0));
```
hanumantmk: Using it this way, you don't even need the impls.  I.e. you could just:

template <typename T>
constexpr std::false_type is_bson_appendable_impl(...);

src/mongo_odm/query_builder.hpp
```diff
@@ -25,93 +26,453 @@
...
+    builder.append(value);
+}
+
+// Specialization for non-iterable, non-expression types that must be serialized.
+template <typename T>
+typename std::enable_if<!is_bson_appendable<T>::value && !is_iterable_not_string<T>::value &&
+                            !(is_sort_expression<T>::value || is_query_expression<T>::value ||
+                              is_update_expression<T>::value),
+                        void>::type
+append_value_to_bson(T value, bsoncxx::builder::core &builder) {
```
hanumantmk: any reason to take these by value instead of const& ?

src/mongo_odm/query_builder.hpp
```diff
@@ -25,93 +26,453 @@
...
+    auto serialized_value = bson_mapper::to_document<T>(value);
+    builder.append(bsoncxx::types::b_document{serialized_value});
+}
+
+// Specialization for iterable types that must be serialized.
+template <typename Iterable>
+typename std::enable_if<is_iterable_not_string<Iterable>::value, void>::type append_value_to_bson(
+    Iterable arr, bsoncxx::builder::core &builder) {
+    builder.open_array();
+    for (auto x : arr) {
```
hanumantmk: const auto& ?

src/mongo_odm/query_builder.hpp
```diff
@@ -25,93 +26,453 @@
...
+    builder.close_array();
+}
+
+// Specialization for expression types that must be serialized.
+template <typename Expression>
+typename std::enable_if<is_sort_expression<Expression>::value ||
+                            is_query_expression<Expression>::value ||
+                            is_update_expression<Expression>::value,
+                        void>::type
+append_value_to_bson(Expression expr, bsoncxx::builder::core &builder) {
```
hanumantmk: Same question about const& here and below?

src/mongo_odm/query_builder.hpp
```diff
@@ -25,93 +26,453 @@
...
-     * @param  selector_type The type of comparison operator, such at gt (>) or ne (!=).
+     * @param  op            The type of comparison operator, such at gt (>) or ne (!=).
+     */
+    constexpr ComparisonExpr(const NvpT &nvp, const U &field, const char *op)
+        : _nvp(nvp), _field(field), _operator(op) {
+    }
+
+    /**
+     * Takes a comparison expression, and creates a new one with a different operator, but the same
+     * value and field.
```
Machyne: Can you add a comment here about the weird regex type that requires this?

src/mongo_odm/test/query_builder.cpp
```diff
@@ -152,7 +419,180 @@ TEST_CASE("Query Builder", "[mongo_odm::query_builder]") {
...
+            REQUIRE(res);
+            REQUIRE((res->arr == std::vector<int>{4, 5, 6}));
+
+            res = Bar::find_one(MONGO_ODM_KEY(Bar::arr).all(std::vector<int>{4, 5, 6, 7}));
+            REQUIRE(!res);
+        }
+
+        SECTION("Test $elem_match operator.", "[mongo_odm::Nvp::elem_match]") {
+            // Test scalar array
+            auto res = Bar::find_one(MONGO_ODM_KEY(Bar::arr).elem_match(
```
Machyne: per in person review, can we make this look like:

```c++
States::find_one(MONGO_ODM_KEY(States::zips)
                     .elem_match(MONGO_ODM_ELEM(States::zips) >= 20000,
                                 MONGO_ODM_ELEM(States::zips) < 25000));
States::find_one(MONGO_ODM_KEY(States::stores)
                     .elem_match(MONGO_ODM_KEY(Store::name) == "wallmart",
                                 MONGO_ODM_KEY(Store::zip) < 25000));
```
for the schema:
```js
{
    zips: [20000, ...],
    stores: [
        {name: "shop", zip: 12345},
        ...
    ]
}
```

src/mongo_odm/query_builder.hpp
```diff
@@ -120,15 +481,53 @@ class NotExpr {
...
+
+   private:
+    const Expr expr;
+};
+
+/**
+ * This represents a list of expressions.
+ */
 template <typename Head, typename Tail>
 class ExpressionList {
```
Machyne: Suggestion, add a non-type template param of an enum to distinguish list type. e.g.
```c++
enum struct expression_list_type {
    k_update,
    k_query,
    k_sort,
    //etc.
}
```

src/mongo_odm/nvp.hpp
rkargon: This fails if I capture parent by const& - the tests fail either because of segfaults or when the NVP's names become corrupted.

src/mongo_odm/query_builder.hpp
```diff
@@ -120,15 +481,53 @@ class NotExpr {
...
    public:
     /**
      * Constructs an expression list.
-     * @param head  The first element in the list
+     * @param head  The first element in the list.
      * @param tail  The remainder of the list
      */
     constexpr ExpressionList(const Head &head, const Tail &tail) : _head(head), _tail(tail) {
```
hanumantmk: I'd consider re-working this into a non-linked list variant type because, in general, variants and direct expansions are cheaper than recursive variants.  We can push it off after this review, but it can make a big difference when instantiating large numbers of templates.

I.e. something like:

```cpp
template <typename Map, typename... Ts, size_t... idxs>
constexpr void tupleForEachImpl(const std::tuple<Ts...>& tup, Map&& map, std::index_sequence<idxs...>) {
    std::initializer_list<int>{(map(idxs, std::get<idxs>(tup)), 0)...};
}

template <typename Map, typename... Ts>
constexpr void tupleForEach(const std::tuple<Ts...>& tup, Map&& map) {
    return tupleForEachImpl(tup, std::forward<Map>(map), std::index_sequence_for<Ts...>());
}

template <typename... Args>
class ExpressionList {
  ExpressionList(const Args& ...args) : storage(std::make_tuple(args...)) {}

  void append_to_bson(builder::core& builder, bool wrap=false) const {
    tupleForEach(storage, [&](const auto& v) { v.append_to_bson(builder, wrap); });
  }

  std::tuple<Args...> storage;
};
```

src/mongo_odm/util.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+template <bool...>
+struct bool_pack;
+
+/**
+ * A templated struct for determining whether a variadic list of boolean conditions is all true.
+ * i.e. a logical AND of a set of conditions.
+ */
+template <bool... bs>
+struct all_true : public std::is_same<bool_pack<bs..., true>, bool_pack<true, bs...>> {};
```
hanumantmk: You're doing enough fancy stuff here that I think it might be worth it to look at some of the C++17 type traits that work on variadics:

http://en.cppreference.com/w/cpp/types/conjunction
http://en.cppreference.com/w/cpp/types/disjunction

src/mongo_odm/util.hpp
```diff
@@ -0,0 +1,153 @@
...
+using std::end;
+
+template <typename T>
+auto is_iterable_impl(int)
+    -> decltype(begin(std::declval<T &>()) !=
+                    end(std::declval<T &>()),  // begin/end and operator !=
+                void(),                        // Handle evil operator ,
+                ++std::declval<decltype(begin(std::declval<T &>())) &>(),  // operator ++
+                void(*begin(std::declval<T &>())),                         // operator*
+                std::true_type{});
```
hanumantmk: Nevermind C++11.  We've gone full expression sfinae based concepts.

which means you might as well go dig out all the _t traits (enable_if_t in particular).  Maybe even stdx in some of the nicer variable templates (is_same_v<T> versus is_same<T>::value for instance)

src/mongo_odm/nvp.hpp
```diff
@@ -74,17 +159,593 @@ struct Nvp {
...
+
+    /**
+     * Creates a query that uses the $bitsAllSet operator to check a numerical field with a bitmask.
+     * $bitsAllSet checks that every bit in the bitmask is set in the field's value.
+     * @param bitmask - A bitmask to pass to the $bitsAllSet operator
+     * @returns A ComparisonExpr representing this query
+     */
+    template <typename U = no_opt_type,
+              typename = typename std::enable_if<std::is_integral<U>::value>::type>
+    constexpr ComparisonValueExpr<NvpT, std::int64_t> bits_all_set(
```
hanumantmk: We should also take BinData masks

src/mongo_odm/nvp.hpp
```diff
@@ -74,17 +159,593 @@ struct Nvp {
...
+     * takes a bit mask (see above)
+     * @param pos1  The first bit position to check
+     * @param pos2  The second bit position to check
+     * @param positions...  Variadic argument containing further bit positions.
+     * @returns A ComparisonExpr representing this query
+     */
+    template <typename U = no_opt_type,
+              typename = typename std::enable_if<std::is_integral<U>::value>::type,
+              typename... Args>
+    constexpr ComparisonValueExpr<NvpT, std::int64_t> bits_all_set(std::int64_t pos1,
```
hanumantmk: This might be better taking an initializer_list of size_t positions.

src/mongo_odm/util.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+template <typename>
+std::false_type is_iterable_impl(...);
+
+template <typename T>
+using is_iterable = decltype(is_iterable_impl<T>(0));
```
hanumantmk: the stl style for this kind of thing is to add a _t suffix for traits you don't have to ::type on

src/mongo_odm/test/query_builder.cpp
```diff
@@ -162,10 +602,456 @@ TEST_CASE("Query builder works with non-ODM class") {
...
+        REQUIRE(res);
+        REQUIRE(res->modified_count() == 1);
+
+        auto bar = Bar::find_one(MONGO_ODM_KEY(Bar::w) == 1212);
+        REQUIRE(bar);
+        REQUIRE((bar->arr == std::vector<int>{4, 5, 6, 7}));
+
+        // push several element at once
+        res = coll.update_one(MONGO_ODM_KEY(Bar::w) == 1212,
+                              MONGO_ODM_KEY(Bar::arr).push(std::vector<int>{7, 8, 9}));
```
hanumantmk: They're just tests, but I'd presume these work as well with initializer lists, which will probably look a little nicer

src/mongo_odm/test/query_builder.cpp
```diff
@@ -162,10 +602,456 @@ TEST_CASE("Query builder works with non-ODM class") {
...
+                                  .sort(MONGO_ODM_KEY(Point::x).sort(false)));
+        REQUIRE(res);
+        REQUIRE(res->modified_count() == 1);
+
+        bar = Bar::find_one(MONGO_ODM_KEY(Bar::w) == 1212);
+        REQUIRE(bar);
+        REQUIRE((bar->pts == std::vector<Point>{{11, 12}, {9, 10}}));
+
+        // push with $position option, to insert new elements at a specific place in the array.
+        res = coll.update_one(MONGO_ODM_KEY(Bar::w) == 1212,
```
hanumantmk: I'm not sure if I'm missing it, but do we have any update_many tests?

src/mongo_odm/nvp.hpp
hanumantmk: Is that true if you also capture the param in the ctor by const& rather than by value?

rkargon: woops, missed that, it works fine when using references in the ctor.

src/mongo_odm/nvp.hpp

src/mongo_odm/nvp.hpp

src/mongo_odm/query_builder.hpp
rkargon: @Machyne it seems like without the `int` argument, the two function definitions are ambiguous and it fails to compile.

src/mongo_odm/query_builder.hpp
rkargon: Sorry, I don't follow.
Would it be possible to do something like this?
`struct is_bson_appendable : public decltype(std::declval<bsoncxx::builder::core>().append(std::declval<T>()),  std::true_type{}) {};`

src/mongo_odm/query_builder.hpp
rkargon: nope, fixed.

src/mongo_odm/query_builder.hpp
rkargon: done.

src/mongo_odm/query_builder.hpp
rkargon: fixed.

src/mongo_odm/query_builder.hpp
rkargon: done.

src/mongo_odm/nvp.hpp
rkargon: Using this pointer causes tests to fail, with what looks like memory alignment issues. I get errors like: 
```
/Users/raphaelkargon/Documents/mongo-cxx-odm/src/mongo_odm/test/query_builder.cpp:122: FAILED:
  REQUIRE( nvp_child.parent.parent.t == &BarParent::b )
with expansion:
  0x000000010fa78e4f == 0x0000000000000000
```
@hanumantmk thoughts?
Matt got the `self` pointer to work in a simplified example, but it fails here.

src/mongo_odm/macros.hpp
```diff
@@ -0,0 +1,95 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <mongo_odm/config/prelude.hpp>
+
+// Preprocessor templates for manipulating multiple arguments.
+#define PP_NARG(...) PP_NARG_(__VA_ARGS__, PP_RSEQ_N())
```
hanumantmk: Let's prefix these macros so they can't conflict with consumer code

src/mongo_odm/nvp.hpp
```diff
@@ -160,13 +94,36 @@ class Nvp : public NvpCRTP<Nvp<Base, T>, T> {
...
     }
 
-    virtual std::string get_name() const {
+    /**
+    * Creates an expression that sets a date value to the current date.
+    * This is only enabled for std::chrono::time/duration values, as well as b_date.
+    */
+    template <typename U = no_opt_type>
+    constexpr typename std::enable_if<is_date<U>::value, CurrentDateExpr<Nvp<Base, T>>>::type
+    operator=(const current_date_t& cd) const {
```
hanumantmk: you can just not name the parameter if you don't want to use it.

I.e.

operator=(const current_date_t&) const { return {*this, true}; }

src/mongo_odm/query_builder.hpp
hanumantmk: I meant that you could leave the definition of is_bson_appendable the same and remove the impl definitions (just leaving the declarations) and it would all still work.

You get truthy or falsey-ness just from the return type, which is determined just from overload selection (not actually running anything).

src/mongo_odm/nvp.hpp
rkargon: How about having and `append_name()` function instead, that takes a BSON builder and appends the various parts of the name one by one? That should get rid of unnecessary construction of strings, correct?

src/mongo_odm/util.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+template <typename T>
+using remove_optional_t = typename remove_optional<T>::type;
+
+constexpr std::int64_t bit_positions_to_mask() {
+    return 0;
+}
+
+template <typename... Args>
+constexpr std::int64_t bit_positions_to_mask(std::int64_t pos, Args... positions) {
```
rkargon: Done.

src/mongo_odm/query_builder.hpp
```diff
@@ -535,30 +531,40 @@ class FreeExpr {
     const Expr expr;
 };
 
+template <typename Map, typename... Ts, size_t... idxs>
+constexpr void tupleForEachImpl(const std::tuple<Ts...> &tup, Map &&map,
```
hanumantmk: Let's migrate this over to the utility header

src/mongo_odm/query_builder.hpp
rkargon: Ah, got it.

src/mongo_odm/nvp.hpp
hanumantmk: The core builder doesn't really work that way at the moment (it wants the key all at once).

There is one way we could skip all the runtime overhead.  If we changed the Nvp's to include a size_t non-type parameter in their template list, and used that to store the key, we could recursively make bigger name buffers for the NvpChild'ren as well, which would allow us to always have a fully formed key on hand as needed.

We'd find that value by writing a constexpr strlen and using that in the macros we use to tag NVPs.

src/mongo_odm/test/query_builder.cpp
rkargon: The compiler has trouble inferring the type of the initializer lists, since they can be any iterable.

src/mongo_odm/nvp.hpp

src/mongo_odm/macros.hpp
rkargon: Done.

src/mongo_odm/macros.hpp
```diff
@@ -0,0 +1,97 @@
...
+#define MONGO_ODM_PP_NARG_(...) MONGO_ODM_PP_ARG_N(__VA_ARGS__)
+#define MONGO_ODM_PP_ARG_N(_0, _1, _2, _3, _4, _5, _6, _7, _8, _9, _10, _11, _12, _13, _14, _15, \
+                           _16, _17, _18, _19, _20, _21, _22, _23, _24, _25, _26, _27, _28, _29, \
+                           _30, _31, _32, _33, _34, _35, _36, _37, _38, _39, _40, _41, _42, _43, \
+                           _44, _45, _46, _47, _48, _49, _50, _51, _52, _53, _54, _55, _56, _57, \
+                           _58, _59, _60, _61, _62, _63, _64, _65, _66, _67, _68, _69, _70, _71, \
+                           _72, _73, _74, _75, _76, _77, _78, _79, _80, _81, _82, _83, _84, _85, \
+                           _86, _87, _88, _89, _90, _91, _92, _93, _94, _95, _96, _97, _98, _99, \
+                           _100, _101, N, ...)                                                   \
+    N
```
Machyne: This is really weird. Did clang-format do this?

rkargon: Yup.

src/mongo_odm/nvp.hpp
```diff
@@ -21,51 +21,76 @@
...
+struct current_date_t {
+    constexpr current_date_t() {
+    }
+};
+constexpr current_date_t current_date;
+
+// Forward declarations for Expression type trait structs
+
+template <typename T>
+struct is_sort_expression;
```
Machyne: Can you move all of these type traits to a consolidated header file and namespace? Either `impl` or `details`.

src/mongo_odm/nvp.hpp
```diff
@@ -74,17 +99,643 @@ struct Nvp {
...
+    constexpr UpdateExpr<Nvp<Base, T>, no_opt_type> operator=(const no_opt_type& val) const {
+        return {*this, val, "$set"};
+    }
+
+    /**
+    * Creates an expression that sets a date value to the current date.
+    * This is only enabled for std::chrono::time/duration values, as well as b_date.
+    */
+    template <typename U = no_opt_type>
+    constexpr typename std::enable_if<is_date<U>::value, CurrentDateExpr<Nvp<Base, T>>>::type
```
Machyne: Can you just say `CurrentDateExpr<Nvp>` here?

src/mongo_odm/query_builder.hpp
```diff
@@ -25,93 +26,462 @@
...
 #include <mongo_odm/nvp.hpp>
+#include <mongo_odm/util.hpp>
 
 namespace mongo_odm {
 MONGO_ODM_INLINE_NAMESPACE_BEGIN
 
-// Forward declarations
-template <typename Base, typename T>
-class NotExpr;
+// Enum of possible expressiokn types.
```
Machyne: "expression"

src/mongo_odm/query_builder.hpp
```diff
@@ -25,93 +26,462 @@
...
+enum struct expression_category_enum { none, update, query, sort };
```
Machyne: just `expression_category`, don't need "enum"

src/mongo_odm/test/query_builder.cpp

src/mongo_odm/nvp.hpp

src/mongo_odm/query_builder.hpp
rkargon: Done.

src/mongo_odm/nvp.hpp
```diff
@@ -74,17 +58,646 @@ struct Nvp {
...
+    */
+    template <typename U = no_opt_type>
+    constexpr typename std::enable_if<is_date<U>::value, CurrentDateExpr<Nvp>>::type operator=(
+        const current_date_t&) const {
+        return {*this, true};
+    }
+
+    /**
+     * Creates an expression that sets a date value to the current date.
+     * This is only enabled for std::chrono::time/duration values, as well as b_date.
```
adamchel: This one is actually only enabled for `bsoncxx::types::b_timestamp` values.

src/mongo_odm/nvp.hpp
```diff
@@ -74,17 +58,646 @@ struct Nvp {
...
+    */
+    template <typename U = no_opt_type>
+    constexpr typename std::enable_if<is_date<U>::value, CurrentDateExpr<NvpChild>>::type operator=(
+        const current_date_t&) const {
+        return {*this, true};
+    }
+
+    /**
+     * Creates an expression that sets a date value to the current date.
+     * This is only enabled for std::chrono::time/duration values, as well as b_date.
```
adamchel: ditto about b_timestamp

src/mongo_odm/nvp.hpp
rkargon: Fixed the comment.

src/mongo_odm/nvp.hpp
rkargon: Fixed.

src/mongo_odm/nvp.hpp
```diff
@@ -74,17 +58,646 @@ struct Nvp {
...
+
+    /**
+     * Creates a comparison expression that represents a query with a $regex operator.
+     * Such a query only works for string fields.
+     * @param regex     The regex to check against.
+     * @param options   Options to pass to the regex.
+     */
+    template <typename U = no_opt_type,
+              typename = typename std::enable_if<is_string<U>::value>::type>
+    constexpr ComparisonValueExpr<NvpT, bsoncxx::types::b_regex> regex(const char* regex,
```
adamchel: I'm not completely sure you should be using `b_regex` here. `b_regex` seems to be one of those weird internal types and it has a different BSON type code than a string. Does the $regex operator expect a BSON string type or a BSON regex type? Thoughts @Machyne @hanumantmk ?

src/mongo_odm/nvp.hpp
```diff
@@ -74,17 +58,646 @@ struct Nvp {
...
+     * that adds a list of value to an array. Further modifiers can be given as arguments to this
+     * function, or by modifying the resulting PushUpdateExpr object.
+     * This is only enabled if the current field is an array.
+
+     * @tparam Iterable An iterable that contains the same value type as this field's array.
+     * @tparam Sort     The type of the given sort expression, if one is given.
+     * @param  iter     The iterable containing values to add.
+     * @param  slice    An optional argument containing the value of the $slice modifier.
+     * @param  sort    An optional argument containing an expression for the $sort modifier.
+     * @param  potision    An optional argument containing the value of the $position modifier.
```
adamchel: potision typo

src/mongo_odm/nvp.hpp
rkargon: ah, did not know that. 

The $regex operator can accept either, but the $not operator is more finnicky. The main reason I'm using b_regex here is that the `$not` operator can't have a `$regex` operator inside of it, so when a user applies `operator!` to a regex expression, it would transform it to a `$not` that directly contains the regex object. 

i.e. `{"z" : {$not: /O+D/i }}` works find in the mongo shell, but `{"z" : {$not: {$regex: /O+D/i} }}` produces an error: `"errmsg": "$not cannot have a regex"`

src/mongo_odm/nvp.hpp
rkargon: woops, fixed.

src/mongo_odm/query_builder.hpp
```diff
@@ -24,94 +25,462 @@
...
+        builder.key_view("$search");
+        builder.append(_search);
+        if (_language) {
+            builder.key_view("$language");
+            builder.append(_language.value());
+        }
+        builder.key_view("$caseSensitive");
+        builder.append(_case_sensitive);
+        builder.key_view("$diacriticSensitive");
+        builder.append(_diacritic_sensitive);
```
adamchel: Since $caseSensitive and $diacriticSensitive are false by default, it would probably save some bandwidth over the wire if you only included them if they're true.

rkargon: Good call, implemented.

Machyne: For optional things like this, we can make them `optional` and offer fluent setters, and then always propagate the values that are explicitly set (even to the default -- which might change) and ignore the unset values.

rkargon: By "fluent", do you mean something like `TextExpression(args).case_sensitive(true).diacritic_sensitive(false)` ?

rkargon: Done.

buildscripts/generate_macros.py
```diff
@@ -0,0 +1,91 @@
...
+    #define MONGO_ODM_CHILD_ELEM(value) MONGO_ODM_KEY(value).element()
+
+    #include <mongo_odm/config/postlude.hpp>
+// clang-format on"""
+    )
+    outfile.write("\n")
+
+if __name__ == "__main__":
+    args = sys.argv[1:]
+    if len(args) < 1 or len(args) > 2:
```
Machyne: `if not (len(args) == 1 or len(args) == 2):`

buildscripts/generate_macros.py
```diff
@@ -0,0 +1,91 @@
...
+            "make_nvp_with_parent(MONGO_ODM_KEY_BY_VALUE(&std::decay_t<"
+            "typename decltype(MONGO_ODM_CHILD%d(base, %s))::type>::field%d), "
+            "MONGO_ODM_CHILD%d(base, %s))") % \
+            (i, all_fields, i-1, prev_fields, i, i-1, prev_fields)
+        outfile.write(s)
+        outfile.write("\n\n")
+
+    outfile.write(
+    """
+    #define MONGO_ODM_CHILD(type, ...) \
```
Machyne: I think `MONGO_ODM_CHILD` and `MONGO_ODM_CHILD_ELEM` should go into macros.hpp and `MONGO_ODM_CHILD1` should be in this one.

buildscripts/generate_macros.py
rkargon: How about `if len(args) not in [1, 2]:`?

src/mongo_odm/expression_syntax.hpp
```diff
@@ -0,0 +1,176 @@
...
+// Enum of possible expression types.
+enum struct expression_category { none, query, update, sort };
+
+// Forward declarations for Expression type trait structs
+template <expression_category, typename>
+struct is_expression_type;
+
+/* Type traits structs for various expressions */
+template <expression_category expr_type>
+struct expression_category_type {
```
Machyne: Can this be `expression_category_t`?

buildscripts/generate_macros.py
rkargon: Done.

src/mongo_odm/expression_syntax.hpp
```diff
@@ -0,0 +1,176 @@
...
+    static constexpr expression_category value = expr_type;
+};
+
+/* forward declarations of expression classes */
+
+template <typename NvpT>
+class SortExpr;
```
Machyne: For naming consistency, we should change these all to `sort_expr` etc. How much extra work is that?

Machyne: It can also be caught in our grand renaming sweep, but I think it's better to do here if possible.

src/mongo_odm/expression_syntax.hpp
```diff
@@ -0,0 +1,176 @@
...
+class PushUpdateExpr;
+
+template <typename NvpT, typename Integer>
+class BitUpdateExpr;
+
+namespace details {
+
+/* type traits for classifying expressions by their category. */
+
+using expression_none_type = expression_category_type<expression_category::none>;
```
Machyne: `expression_none_t` etc

src/mongo_odm/expression_syntax.hpp
rkargon: This shouldn't be too hard, I can do it here.

src/mongo_odm/expression_syntax.hpp
```diff
@@ -0,0 +1,176 @@
...
+struct expression_type<AddToSetUpdateExpr<NvpT, U>> : public expression_update_type {};
+
+template <typename NvpT, typename U, typename Sort>
+struct expression_type<PushUpdateExpr<NvpT, U, Sort>> : public expression_update_type {};
+
+template <typename NvpT, typename Integer>
+struct expression_type<BitUpdateExpr<NvpT, Integer>> : public expression_update_type {};
+
+template <expression_category type, typename T>
+struct is_expression_type {
```
Machyne: What do you think about the name `is_expression_of_type`?

src/mongo_odm/macros.hpp
```diff
@@ -0,0 +1,97 @@
...
+// limitations under the License.
+
+#pragma once
+
+#include <mongo_odm/config/prelude.hpp>
+
+// Preprocessor templates for manipulating multiple arguments.
+#define MONGO_ODM_PP_NARG(...) MONGO_ODM_PP_NARG_(__VA_ARGS__, MONGO_ODM_PP_RSEQ_N())
+#define MONGO_ODM_PP_NARG_(...) MONGO_ODM_PP_ARG_N(__VA_ARGS__)
+#define MONGO_ODM_PP_ARG_N(_0, _1, _2, _3, _4, _5, _6, _7, _8, _9, _10, _11, _12, _13, _14, _15, \
```
Machyne: Is this used anywhere other than for `ODM_CHILD{{N}}`? If not, what do you think about putting these `PP_NARG`s marcros into that generated file?

Machyne: *that way we can generate the proper number in all cases

src/mongo_odm/nvp.hpp
```diff
@@ -21,51 +21,35 @@
...
+
+#include <mongo_odm/expression_syntax.hpp>
+#include <mongo_odm/macros.hpp>
+#include <mongo_odm/util.hpp>
 
 namespace mongo_odm {
 MONGO_ODM_INLINE_NAMESPACE_BEGIN
 
+template <typename NvpT, typename T>
+class NvpCRTP;
```
Machyne: `nvp_base`?

src/mongo_odm/nvp.hpp
```diff
@@ -21,51 +21,35 @@
...
+};
+constexpr current_date_t current_date;
+
 /**
  * An object that represents a name-value pair of a member in an object.
  * It is templated on the class of the member and its type.
  */
 template <typename Base, typename T>
-struct Nvp {
+class Nvp : public NvpCRTP<Nvp<Base, T>, T> {
```
Machyne: I think it'd be good to switch the names here. Can you change this to `class nvp`?

src/mongo_odm/nvp.hpp
Machyne: Did something actually change here?

src/mongo_odm/nvp.hpp
Machyne: I think using `b_regex` here is good. You might also consider adding a `b_regex` overload. I think it's a pretty common type, and at the very least I know a lot of users do use this.

src/mongo_odm/nvp.hpp
```diff
@@ -74,17 +58,646 @@ struct Nvp {
...
+    }
 };
 
+/**
+ * A type trait struct that inherits from std::true_type
+ * if the given type parameter is a name-value pair,
+ * and from std::false_type otherwise.
+ */
+template <typename>
+struct is_nvp_type : public std::false_type {};
```
Machyne: `is_nvp`

src/mongo_odm/nvp.hpp
```diff
@@ -74,17 +58,646 @@ struct Nvp {
...
+
+template <typename Base, typename T>
+struct is_nvp_type<Nvp<Base, T>> : public std::true_type {};
+
+template <typename Base, typename T, typename Parent>
+struct is_nvp_type<NvpChild<Base, T, Parent>> : public std::true_type {};
+
+template <typename T>
+struct is_nvp_type<FreeNvp<T>> : public std::true_type {};
+
```
Machyne: maybe add a `using is_nvp_v = is_nvp<T>`. I think the`_v` idiom is getting more common

src/mongo_odm/expression_syntax.hpp
rkargon: Done.

src/mongo_odm/query_builder.hpp
```diff
@@ -24,94 +25,482 @@
...
+}
+
+/**
+ * An expression that represents a sorting order.
+ * This consists of a name-value pair and a boolean specifying ascending or descending sort
+ * order.
+ * The resulting BSON is {field: +/-1}, where +/-1 corresponds to the sort order.
+ */
+template <typename NvpT>
+class SortExpr {
```
Machyne: `class sort_expr` ditto for other expr's

src/mongo_odm/nvp.hpp
Machyne: *oopts, I mean `constexpr bool is_nvp_v`

rkargon: Do you mean `constexpr bool is_nvp_v = is_nvp<T>::value`?

buildscripts/generate_macros.py
rkargon: Done.

src/mongo_odm/mongo_odm_child_autogen.hpp
```diff
@@ -17,9 +17,16 @@
```
Machyne: I think we propbably don't want this file in version control? If it's auto-generated, is there a reason not to gitignore it?

buildscripts/generate_macros.py
Machyne: `if len(args) not in [1, 2]:` sounds good!

src/mongo_odm/util.hpp
```diff
@@ -0,0 +1,170 @@
...
+                void(),                        // Handle evil operator ,
+                ++std::declval<decltype(begin(std::declval<T &>())) &>(),  // operator ++
+                void(*begin(std::declval<T &>())),                         // operator*
+                std::true_type{});
+
+template <typename>
+std::false_type is_iterable_impl(...);
+
+template <typename T>
+using is_iterable_t = decltype(is_iterable_impl<T>(0));
```
Machyne: Can you change this to `is_iterable` and then add an `is_iterable_v`? ditto for all type traits in here. Let's just provide everything :P

src/mongo_odm/nvp.hpp
```diff
@@ -21,67 +21,683 @@
...
+    constexpr update_expr<nvp<Base, T>, no_opt_type> operator=(const no_opt_type& val) const {
+        return {*this, val, "$set"};
+    }
+
+    /**
+    * Creates an expression that sets a date value to the current date.
+    * This is only enabled for std::chrono::time/duration values, as well as b_date.
+    */
+    template <typename U = no_opt_type>
+    constexpr typename std::enable_if<is_date<U>::value, current_date_expr<nvp>>::type operator=(
```
Machyne: Since we're full on 14, can you change all of the `enable_if<FOO>::type` to `enable_if_t<FOO>`?

src/mongo_odm/mongo_odm_child_autogen.hpp
rkargon: Done.

src/mongo_odm/util.hpp
rkargon: Done.

src/mongo_odm/nvp.hpp

src/mongo_odm/nvp.hpp
```diff
@@ -21,67 +21,686 @@
...
+    constexpr update_expr<nvp<Base, T>, no_opt_type> operator=(const no_opt_type& val) const {
+        return {*this, val, "$set"};
+    }
+
+    /**
+    * Creates an expression that sets a date value to the current date.
+    * This is only enabled for std::chrono::time/duration values, as well as b_date.
+    */
+    template <typename U = no_opt_type>
+    constexpr typename std::enable_if_t<is_date_v<U>, current_date_expr<nvp>> operator=(
```
hanumantmk: You don't need typename for enable_if_t (that's the nice part of the _t).

Here and elsewhere

src/mongo_odm/query_builder.hpp
```diff
@@ -24,94 +25,483 @@
...
+ * passes the value directly to the builder. If it cannot be nicely appended, it is first
+ * serialized
+ * and then added as a sub-document to the builder. If the value is a container/iterable, it is
+ * serialized into a BSON array. If the value is a query builder epxression, it is serialized
+ * using
+ * its member function .append_to_bson(builder).
+ */
+// Specialization for appendable types.
+template <typename T>
+typename std::enable_if_t<is_bson_appendable_v<T>, void> append_value_to_bson(
```
hanumantmk: the default type of enable_if_t is void.  That's a param you can omit

src/mongo_odm/query_builder.hpp
```diff
@@ -24,94 +25,483 @@
...
+     * The resulting BSON is of the form "{ key: { $mod: [ divisor, remainder ] } }".
+     */
+    operator bsoncxx::document::view_or_value() const {
+        auto builder = bsoncxx::builder::core(false);
+        append_to_bson(builder);
+        return builder.extract_document();
+    }
+
+   private:
+    const NvpT _nvp;
```
hanumantmk: const& for these?

src/mongo_odm/query_builder.hpp
```diff
@@ -207,139 +612,810 @@ class BooleanExpr {
...
+     * Creates a BSON value of the form "$op: {field: value}"
+     */
+    operator bsoncxx::document::view_or_value() const {
+        auto builder = bsoncxx::builder::core(false);
+        append_to_bson(builder);
+        return {builder.extract_document()};
+    }
+
+   private:
+    NvpT _nvp;
```
hanumantmk: should this be a mutable value?

I'm a little confused by the various ways we capture our Nvp

hanumantmk:
per in person conversation.  After the last couple of nits: lgtm

src/mongo_odm/query_builder.hpp
rkargon: This segfaults, per our in-person conversation.

rkargon:
@hanumantmk nits have been fixed. @Machyne any comments?

Machyne:
🚢 !!

PR #19, opened 2016-07-27
CXXODM-42 Add operator[] to query builder, along with the $ update operator. 

rkargon:
The query `{"user.scores.4": 100}` can now be expressed as `MONGO_ODM_KEY(User::scores)[4] == 100`. 

The query `{$set: {"user.scores.$": 100}}` can now be expressed as `MONGO_ODM_KEY(User::scores).update_first() = 100`.

### DIFF
--- a/src/mongo_odm/nvp.hpp
+++ b/src/mongo_odm/nvp.hpp
@@ -172,6 +172,48 @@ class nvp_child : public nvp_base<nvp_child<Base, T, Parent>, T> {
     const Parent& parent;
 };
 
+template <typename NvpT>
+class array_element_nvp
+    : public nvp_base<array_element_nvp<NvpT>, iterable_value_t<typename NvpT::no_opt_type>> {
+   public:
+    using type = iterable_value_t<typename NvpT::no_opt_type>;
+    // In case this field is wrapped in an optional, store the underlying type.
+    using no_opt_type = remove_optional_t<type>;
+
+    constexpr array_element_nvp(const NvpT& nvp, std::size_t i) : _nvp(nvp), _i(i) {
+    }
+
+    /**
+     * Creates an update expression that sets the field to the given value.
+     */
+    constexpr update_expr<array_element_nvp<NvpT>, no_opt_type> operator=(
+        const no_opt_type& val) const {
+        return {*this, val, "$set"};
+    }
+
+    /**
+     * Returns the qualified name of this field in dot notation, i.e. "array.i".
+     * @return A string containing the name of this field in dot notation.
+     */
+    std::string get_name() const {
+        std::string s;
+        return append_name(s);
+    }
+
+    /**
+     * Returns the qualified name of this field in dot notation, i.e. "field.<index>", where index
+     * is a number.
+     * @return A string containing the name of this field in dot notation.
+     */
+    std::string& append_name(std::string& s) const {
+        return _nvp.append_name(s).append(1, '.').append(std::to_string(_i));
+    }
+
+   private:
+    const NvpT& _nvp;
+    const std::size_t _i;
+};
+
 /**
  * Represents a field that does not have a name, i.e. an array element.
  * This is used in $elemMatch expressions on scalar arrays, where the elements can be compared but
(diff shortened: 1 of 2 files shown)


src/mongo_odm/nvp.hpp
```diff
@@ -638,11 +732,11 @@ class nvp_base {
      * @param  sort    An optional argument containing an expression for the $sort modifier.
      * @param  position    An optional argument containing the value of the $position modifier.
      */
-    template <typename U = no_opt_type, typename = std::enable_if_t<is_iterable_not_string_v<U>>,
-              typename Iterable, typename = enable_if_matching_iterable_t<Iterable>,
-              typename Sort = int,
-              typename = std::enable_if_t<details::is_sort_expression_v<Sort> ||
-                                          std::is_same<int, Sort>::value>>
+    template <
+        typename U = no_opt_type, typename = std::enable_if_t<is_iterable_not_string_v<U>>,
```
Machyne: Seems like this change is formatting only? I think clang-format was green before. What happened?

src/mongo_odm/test/query_builder.cpp
```diff
@@ -127,6 +128,7 @@ TEST_CASE("Test nested member access.", "[mongo_odm::nvp_child]") {
     }
 
     SECTION("Nested member access using MONGO_ODM_CHILD macro") {
+        // Nvp's must be used as temporary objects.
```
Machyne: Why is this?

src/mongo_odm/nvp.hpp
rkargon: I think my editor has the wrong version of clang-format, I'll fix this.

src/mongo_odm/test/query_builder.cpp
rkargon: The NvpChild object stores parent Nvp's by reference, and the other expressions / nvp types do this as well. Which means that if one is stored in a variable, and then used later on, it's `Parent` member is now invalid. IIRC @hanumantmk  mentioned that if we store everything by value, when the expressions get built, successive copies of sub-expressions might end up blowing the stack. 

It might be worth having the Nvp types store their members by value, since they won't get so big, and just leaving the expressions to store members by reference which would make these tests clearer.

src/mongo_odm/nvp.hpp
```diff
@@ -541,6 +630,11 @@ class nvp_base {
     }
 
     /* Array update operators */
+    template <typename U = no_opt_type, typename = std::enable_if_t<is_iterable_not_string_v<U>>>
+    constexpr dollar_operator_nvp<NvpT> update_first() const {
```
adamchel: Would it perhaps be more clear if this was called `update_first_match()`?

src/mongo_odm/test/query_builder.cpp
```diff
@@ -703,77 +748,90 @@ TEST_CASE("Update Builder", "mongo_odm::update_expr") {
...
             REQUIRE(bar->p.x == initial_p_x - 1);
+            REQUIRE(bar->arr[2] == initial_arr_2 - 1);
         }
 
         {
             // Test operator+=
             initial_x1 = bar->x1;
             initial_x2 = bar->x2.value();
             initial_p_x = bar->p.x;
+            initial_arr_2 = bar->arr[2];
```
Machyne: Did you mean to also update this value then `REQUIRE` that it's 5 more?

src/mongo_odm/test/query_builder.cpp
```diff
@@ -791,6 +849,7 @@ TEST_CASE("Update Builder", "mongo_odm::update_expr") {
...
```
Machyne: ditto

src/mongo_odm/test/query_builder.cpp
```diff
@@ -808,6 +867,7 @@ TEST_CASE("Update Builder", "mongo_odm::update_expr") {
...
```
Machyne: ditto

src/mongo_odm/test/query_builder.cpp
```diff
@@ -961,6 +1021,19 @@ TEST_CASE("Update Builder", "mongo_odm::update_expr") {
 
     /* Array update operators */
 
+    SECTION("Test $ array update operator.", "[mongo_odm::dollar_operator_nvp]") {
+        auto res =
+            coll.update_one((MONGO_ODM_KEY(Bar::w) == 555,
+                             MONGO_ODM_KEY(Bar::arr).elem_match(MONGO_ODM_ELEM(Bar::arr) == 5)),
```
Machyne: I don't think you need `elem_match` for single equality queries.

src/mongo_odm/test/query_builder.cpp
```diff
@@ -961,6 +1021,19 @@ TEST_CASE("Update Builder", "mongo_odm::update_expr") {
...
+                            MONGO_ODM_KEY(Bar::arr).update_first() = 500);
```
Machyne: Thoughts on renaming this to `first_match()`? `update_first` is a little confusing.

src/mongo_odm/nvp.hpp
Machyne: I had the same thought, but I'm in favor of `first_match()`

rkargon: `first_match` sounds good, since it returns a name-value pair (a "noun") that gets modified later when it's put into an update expression.

src/mongo_odm/test/query_builder.cpp
rkargon: yup, done.

src/mongo_odm/test/query_builder.cpp
rkargon: Currently, without `elem_match`, the query becomes `{w: {$eq: 555}, $eq: 5}`, which is syntactically correct. 
One way to solve this would be to allow something like `MONGO_ODM_KEY(Bar::arr) == 5`, which would be comes `{arr: {$eq: 5}}`. Right now this is disabled because array fields must be compared to array values, but I can put it a ticket to allow arrays to be directly compared to a value which is the type of their element. So `Bar::arr`, a vector of `int`s, could be compared to an `int` value like `5`. Which is syntactically valid for Mongo queries, I just didn't realize that was possible at the time.

src/mongo_odm/test/query_builder.cpp

src/mongo_odm/test/query_builder.cpp
Machyne: `MONGO_ODM_KEY(Bar::arr) == 5` is what I was thinking. Sounds good!

Machyne:
LGTM pending green build

src/mongo_odm/test/query_builder.cpp
rkargon: Done, issue added as `CXXODM-55`.

PR #27, opened 2016-08-01
CXXODM-55 Allow arrays to be directly compared to scalar values in query builder

rkargon:
The following syntax is now valid:

`Foo::find_one(MANGROVE_KEY(nums_arr) > 7)`

and is "equivalent" to:

`Foo::find_one(MANGROVE_KEY(nums_arr).elem_match(MANGROVE_ELEM(nums_arr) > 7))`

(i.e. The syntax is equivalent when specifying only **one** condition, but with multiple conditions there are some interesting nuances. However, it still *mirrors MongoDB query semantics*.

### DIFF
--- a/hugo/content/3-queries/operators.md
+++ b/hugo/content/3-queries/operators.md
@@ -17,6 +17,12 @@ They work on any field type, and the right-hand value must be the same type as t
 Free functions are also provided in the `mangrove` namespace that can be used in lieu of the overloaded
 operators.
 
+{{% notice tip %}}
+When working with array fields, one can also compare them to values that are the type of the array's elements.
+This provides similar functionality to using the `$elemMatch` operator, with a few caveats.
+For more information on this, and an example, see the Mangrove documentation for `$elemMatch`, under the [Array Operators]({{< relref "#array-operators" >}}).
+{{% /notice %}}
+
 - `{{% a_blank "$eq" "https://docs.mongodb.com/manual/reference/operator/query/eq/" %}}`
  ```cpp
  // { age: {$eq: 21} }
@@ -212,7 +218,7 @@ For the following code samples, assume that a User has a field `scores`, which c
     ```
 
 - `{{% a_blank "$elemMatch" "https://docs.mongodb.com/manual/reference/operator/elemMatch/" %}}`
-
+    
     This operator accepts a query, and returns documents if the elements inside a field's array match
     the given queries. The queries can be constructed in Mangrove using the same syntax as top-level queries.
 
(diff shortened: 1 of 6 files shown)


hugo/content/3-queries/operators.md
```diff
@@ -17,6 +17,12 @@ They work on any field type, and the right-hand value must be the same type as t
 Free functions are also provided in the `mangrove` namespace that can be used in lieu of the overloaded
 operators.
 
+{{% notice tip %}}
+When working with array fields, one can also compare them to values that are the type of the array's elements.
+This provides the same functionality as using the `$elemMatch` operator.
```
Machyne: There is a subtle difference here between elem match and the raw operators. For example, if I have some documents like this:
```js
{
  a: [ { x: 4, y: 1 }, { x: 0, y: 4 } ]
}
{
  a: [ { x: 0, y: 1 }, { x: 4, y: 4 } ]
}
```

Then this can happen:
```
> db.test.find({ 'a.x': 0, 'a.y': 1 })
{ "_id" : ObjectId("579f984e454ad917bc9f6025"), "a" : [ { "x" : 4, "y" : 1 }, { "x" : 0, "y" : 4 } ] }
{ "_id" : ObjectId("579f984e454ad917bc9f6026"), "a" : [ { "x" : 0, "y" : 1 }, { "x" : 4, "y" : 4 } ] }

> db.test.find({ 'a': { '$elemMatch': { 'x': 0, 'y': 1 } } })
{ "_id" : ObjectId("579f984e454ad917bc9f6026"), "a" : [ { "x" : 0, "y" : 1 }, { "x" : 4, "y" : 4 } ] }
```

elemMatch has to match with a single element, whereas just comparing to a scalar is done individually and will return documents where any element matches.

hugo/content/3-queries/operators.md
```diff
@@ -212,7 +218,7 @@ For the following code samples, assume that a User has a field `scores`, which c
     ```
 
 - `{{% a_blank "$elemMatch" "https://docs.mongodb.com/manual/reference/operator/elemMatch/" %}}`
-
+    <a name="elemMatch"></a>
```
Machyne: What is this tag doing here?

rkargon: This creates an anchor, so that you can link to it from the same page. At the top of the page there's a part that says "...see the Mangrove documentation on __elemMatch__ ..." and there's a link to this part of the documentation. 

@adamchel do you know a better way to do this in Markdown / Hugo? AFAIK Hugo only provides anchors for headings (created using `#`), so for list items you have to add these anchors manually.

src/mangrove/query_builder.hpp
```diff
@@ -1132,79 +1131,169 @@ class bit_update_expr {
...
+
 template <typename NvpT, typename = std::enable_if_t<is_nvp_v<NvpT>>>
 constexpr comparison_expr<NvpT, typename NvpT::no_opt_type> gt(
     const NvpT &lhs, const typename NvpT::no_opt_type &rhs) {
     return {lhs, rhs, "$gt"};
 }
 
+template <typename NvpT, typename = std::enable_if_t<is_nvp_v<NvpT>>,
+          typename = std::enable_if_t<is_iterable_v<typename NvpT::no_opt_type>>>
+constexpr comparison_expr<NvpT, typename NvpT::array_element_type> gt(
```
Machyne: I think these can go away using [std::conditional](http://en.cppreference.com/w/cpp/types/conditional)
Something like this(?):

```c++
template <
    typename NvpT,
    typename T = std::conditional_t<
        is_nvp_v<NvpT>,
        NvpT::no_opt_type,
        std::enable_if_t<is_iterable_v<typename NvpT::no_opt_type>, NvpT::array_element_type>>>
constexpr comparison_expr<NvpT, T> gt(
    const NvpT& lhs, const T& rhs) {
    return {lhs, rhs, "$gt"};
}
```

rkargon: It's awkward, because you actually want both no_opt_type and array_element_type to be allowed. Another level of confusion is that use std::is_same on T will cause issues, since passing string literals (const char[]) to std::string fields will fail is_same. std::is_convertible also is problematic because passing `int32` to an `int64` field will be pass is_convertible, but then end up inserting an int32 into what should be an int64, and cause de-serialization errors. 

I might be able to do something like:
```
template <typename NvpT, typename U,
          typename = std::enable_if_t<is_nvp_v<NvpT>>,
          typename R = std::conditional<is_convertible<U, array_element_type>, array_element_type,
                                        no_opt_type>>
constexpr comparison_expr<NvpT, R> gt(const NvpT& lhs, const U& rhs);
```

plus or minus some `_v`'s and `::value`'s here and there. 
The idea is that the first enable_if restricts the type of U loosely, using `is_convertible`, but then `R` is chosen to be the appropriate Nvp type, so that the comparison_expr knows what type it's dealing with. (and doesn't accidentally send int32's to int64's and etc. )

rkargon: This causes issues when I use call function with an `int32_t` and a field of type `int64_t`. It ends up storing `0`. I think this might have to do with the fact that comparison_expr stores fields by reference.

rkargon: As per conversation w/ Matt, I'll keep two signatures, one for "same as field type" and one for "array element type", only enabled for iterables.

hugo/content/3-queries/operators.md
adamchel: This page talks about how to do cross references in Hugo: 
https://gohugo.io/extras/crossreferences/

You're not supposed to use pure HTML in Hugo content so probably look into that.

hanumantmk:
Lgtm

PR #29, opened 2016-08-04
CXXODM-65 Refactor $mod operator in query builder to remove mod_expr class

rkargon:
Right now the `$mod` operator is represented by the `mod_expr<nvp>` class, which takes two values (divisor and remainder) and puts then into a BSON array when creating the actual expression. 

I simply replaced this with `comparison_value_expr<nvp, std::array<N, 2>>`, where N is some integral type. Then the values for divisor and remainder get automatically put into an array by the `append_to_bson` function in `comparison_value_expr`.


hanumantmk:
Clever. Lgtm

PR #30, opened 2016-08-04
CXXODM-19 Add support for write concern in active record save() and delete()

adamchel:


hanumantmk:
I wonder if optional options isn't a better approach. This'll get a little awkward after we add another parameter

src/mangrove/model.hpp
```diff
@@ -294,22 +301,27 @@ class model {
...
+        mongocxx::options::update options = mongocxx::options::update()) {
         auto id_match_filter = bsoncxx::builder::stream::document{}
                                << "_id" << this->_id << bsoncxx::builder::stream::finalize;
 
         auto update = bsoncxx::builder::stream::document{}
                       << "$set" << boson::to_dotted_notation_document(*static_cast<T*>(this))
                       << bsoncxx::builder::stream::finalize;
 
-        mongocxx::options::update opts{};
+        options.upsert(true);
```
Machyne: Is there any reason that a user might explicitly set this to false? If so, I think we should honor that. Maybe something like:
```c++
options.upsert(options.upsert().value_or(true));
```

Machyne:
One small question, but LGTM.

src/mangrove/model.hpp
adamchel: If upsert is set to false, it would mean that the object would only be saved if it already exists in the database. I don't really know if that makes sense given the active-record pattern? I would err on the side of keeping it as is, and just documenting that upsert will always be true regardless of what the user sets. thoughts @hanumantmk ?

adamchel: I ended up just documenting that upsert is always true.

PR #31, opened 2016-08-04
CXXODM-58 Add support for `$where` operator in builder.

rkargon:
The builder now supports syntax of the form:
`auto results = User::find(mangrove::where("this.prev_score == this.high_score"));`

The `b_code` and `b_codewscope` types are also supported.


hanumantmk:
Lgtm

PR #33, opened 2016-08-19
CXXODM-61 Write a docs chapter about the BSON mapper

adamchel:


hugo/content/5-the-bson-mapper/archives.md
```diff
@@ -0,0 +1,123 @@
...
+}
+```
+
+The `BSONInputArchive` has the same behavior. When populating a C++ object at the root level, it expects a BSON document with elements for each of its fields. When populating a value, it expects a document with a single element.
+
+## Naming Requirements
+
+Unlike other Cereal archives, the BSON archives require that most values are given a name (either using `CEREAL_NVP`, or `cereal::make_nvp`). The only values that don't need a name when passed into a BSON archive are C++ objects passed in at the root level.
+
+Regular non-object values passed in at the root level still need a name because the the element in the single-element BSON document needs a name. C++ objects serialized as subdocuments also need to have a name, because subdocuments are normal named BSON elements.
```
hanumantmk: "the the element"

hanumantmk:
I feel like I'm missing the link in how this relates to models and how nested documents and arrays work.  This is a good overview of the wrapper, but doesn't offer a practical example

Issue #34, opened 2016-09-30
Build error with mongocxx r3.0.2


gncs:
When building mangrove on top of mongo-c-1.3.5 and mongocxx-r3.0.2 (c3408b6) the following error message appears:

```
/home/gncs/build/mangrove/src/boson/bson_streambuf.cpp: In member function 'int boson::v_noabi::bson_output_streambuf::insert(int)':
/home/gncs/build/mangrove/src/boson/bson_streambuf.cpp:71:37: error: no match for call to '(boson::v_noabi::bson_output_streambuf::document_callback {aka std::function<void(bsoncxx::v_noabi::document::value)>}) (<brace-enclosed initializer list>)'
         _cb({std::move(_data), _len});
                                     ^
In file included from /data/soft/compiler/gcc/gcc-5.4.0-obj/include/c++/5.4.0/memory:79:0,
                 from /home/gncs/local/mongodb-c++-gcc-5.4.0_2/include/bsoncxx/v_noabi/bsoncxx/builder/core.hpp:17,
                 from /home/gncs/local/mongodb-c++-gcc-5.4.0_2/include/bsoncxx/v_noabi/bsoncxx/builder/stream/document.hpp:17,
                 from /home/gncs/build/mangrove/src/boson/bson_streambuf.hpp:23,
                 from /home/gncs/build/mangrove/src/boson/bson_streambuf.cpp:25:
/data/soft/compiler/gcc/gcc-5.4.0-obj/include/c++/5.4.0/functional:2262:5: note: candidate: _Res std::function<_Res(_ArgTypes ...)>::operator()(_ArgTypes ...) const [with _Res = void; _ArgTypes = {bsoncxx::v_noabi::document::value}]
     function<_Res(_ArgTypes...)>::
     ^
/data/soft/compiler/gcc/gcc-5.4.0-obj/include/c++/5.4.0/functional:2262:5: note:   no known conversion for argument 1 from '<brace-enclosed initializer list>' to 'bsoncxx::v_noabi::document::value'
/home/gncs/build/mangrove/src/boson/bson_streambuf.cpp: In member function 'int boson::v_noabi::bson_output_streambuf::insert(int)':
/home/gncs/build/mangrove/src/boson/bson_streambuf.cpp:71:37: error: no match for call to '(boson::v_noabi::bson_output_streambuf::document_callback {aka std::function<void(bsoncxx::v_noabi::document::value)>}) (<brace-enclosed initializer list>)'
[...]

PR #35, opened 2017-01-23
Fix mongocxx r3.0.2 build error

adamchel:
Resolves issue #34 which came about due to a change in pointer types in the C++ driver (https://github.com/mongodb/mongo-cxx-driver/commit/4e49da57b359287d9aa70718d7b025f62fa21268).


CMakeLists.txt
```diff
@@ -84,11 +84,11 @@ set(THIRD_PARTY_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/src/third_party)
 
 enable_testing()
 
-set(LIBBSONCXX_REQUIRED_VERSION 3.0.1)
+set(LIBBSONCXX_REQUIRED_VERSION 3.0.2)
```
Machyne: Should this be `3.1.1`?

src/boson/bson_streambuf.cpp
```diff
@@ -57,8 +57,8 @@ int bson_output_streambuf::insert(int ch) {
         if (_len > BSON_MAX_SIZE) {
             throw std::invalid_argument("BSON document length is too large.");
         }
-        _data = std::unique_ptr<uint8_t, void (*)(std::uint8_t *)>(new uint8_t[_len],
-                                                                   [](uint8_t *p) { delete[] p; });
+        _data = std::unique_ptr<uint8_t[], void (*)(std::uint8_t *)>(
```
Machyne: Does this compile with the default deleter type? i.e. can we construct a `document::value` from a `std::unique_ptr<uint8_t[]>`?

Machyne (approved):
=)

acmorrow:
@adamchel - There have been some recent inquiries about getting this resolved:

https://stackoverflow.com/questions/45430300/which-mongodb-odm-for-c/45449656?noredirect=1#comment78580654_45449656

Can we try to get it moving again?

rkargon:
I'll see if I can get Mangrove running on the latest version of the mongocxx driver, since it looks like we haven't really tried since February. 

In the meantime, it might be worth recommending that the StackOverflow user try to build Mangrove from Adam's fork and seeing if the problem persists.

PR #43, opened 2018-09-14
Fix mangrove crash

aospan:
Crash observed when document length is 257 byte, for example.
It represents as '0x01 01 00 00'.
In this case '_bytes_read == _len' equal to 1 and callback is called
with partial buffer (only 1 byte actually). This cause later crash in
'to_dotted_notation_document'

### DIFF
--- a/src/boson/bson_streambuf.cpp
+++ b/src/boson/bson_streambuf.cpp
@@ -67,7 +67,7 @@ int bson_output_streambuf::insert(int ch) {
     }
 
     // This creates the document from the given bytes, and calls the user-provided callback.
-    if (_bytes_read == _len) {
+    if (_bytes_read == _len && _len > 4) {
         _cb({std::move(_data), _len});
         _bytes_read = 0;
         _len = 0;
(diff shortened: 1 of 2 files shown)


src/boson/bson_streambuf.cpp
```diff
@@ -67,7 +67,7 @@ int bson_output_streambuf::insert(int ch) {
     }
 
     // This creates the document from the given bytes, and calls the user-provided callback.
-    if (_bytes_read == _len) {
+    if (_bytes_read == _len && _len > 4) {
```
rkargon, line 70: How about earlier(e..g line 54)  just returning early if _bytes_read < 4, so we're never dealing with an invalid _len in the later code?

rkargon:
Wow, good catch!

aospan:
@rkargon tnx :)

BTW, better to implement:
streamsize xsputn (const char* s, streamsize n);

then we shouldn't care about buffer assembly and avoid byte-by-byte copy (which can slow down whole process)

PR #44, opened 2018-12-13
Fix embedded doc exception with new mongo-cxx

aospan:
With mongo-cxx-3.2+ we have exception:

terminate called after throwing an instance of
'bsoncxx::v_noabi::exception'
  what():  can't convert builder to a valid view: unmatched key

while using embedded documents. For example:
    {
        "name" : "Jenny",
        "contact_info" :
            {
                "type" : "home"
            }
    }

we have called twice:
 1. for "contact_info" key
 2. for "contact_info.type" key

in mongo-cxx-3.2+ we can't call key_view/key_owned twice. Otherwise we
receive exception as described above.


aospan:
Please ignore this PR. It fixed exception but introduced another problem. I will rework it.

PR #45, opened 2018-12-14
Fix embedded doc exception with new mongo-cxx

aospan:
With modern mongo-cxx (tested on 3.3.1) we have exception while using
embedded documents:

terminate called after throwing an instance of
'bsoncxx::v_noabi::exception'
what(): can't convert builder to a valid view: unmatched key

For example:
{
    "name" : "Jenny",
    "contact_info" :
    {
        "type" : "home"
    }
}

we have called twice:
    1. for "contact_info" key
    2. for "contact_info.type" key

we can't call key_view/key_owned twice.
Otherwise we receive exception as described above.


aospan:
sorry, will version 3 of this patch ! :) this one introduced new problem too.

